PR #25, opened 2018-12-17
feat(js): upgrade Getting Started to v3

francoischalifour:
This updates the InstantSearch.js Getting Started to version 3, using the [beta version](https://cdn.jsdelivr.net/npm/instantsearch.js@beta) for now.

Demo: https://codesandbox.io/s/github/algolia/doc-code-samples/tree/csb-1545053595649/InstantSearch.js/getting-started

_(updated with CodeSandbox)_

### DIFF
--- a/InstantSearch.js/getting-started/package.json
+++ b/InstantSearch.js/getting-started/package.json
@@ -9,6 +9,7 @@
     "lint:fix": "npm run lint -- --fix"
   },
   "devDependencies": {
+    "@types/algoliasearch": "3.30.1",
     "babel-eslint": "8.2.3",
     "eslint": "4.19.1",
     "eslint-config-algolia": "13.1.0",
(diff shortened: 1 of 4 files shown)


InstantSearch.js/getting-started/package.json
```diff
@@ -9,6 +9,7 @@
     "lint:fix": "npm run lint -- --fix"
   },
   "devDependencies": {
+    "@types/algoliasearch": "3.30.1",
```
Haroenv: is this really necessary? VS Code should pick those up automatically 😄

InstantSearch.js/getting-started/src/app.js
```diff
@@ -1,8 +1,8 @@
 /* global instantsearch */
+import algoliasearch from 'algoliasearch/lite';
```
Haroenv: if InstantSearch is global, this should probably be too?

InstantSearch.js/getting-started/package.json
francoischalifour: That was added by CodeSandbox automatically. We can remove it.

InstantSearch.js/getting-started/src/app.js
francoischalifour: Yep!

PR #29, opened 2018-12-17
feat(js): update refresh cache periodically example

francoischalifour:
Demo: https://codesandbox.io/s/github/algolia/doc-code-samples/tree/js-cache-periodically/InstantSearch.js/refresh-cache-periodically

### DIFF
--- a/InstantSearch.js/refresh-cache-periodically/src/app.js
+++ b/InstantSearch.js/refresh-cache-periodically/src/app.js
@@ -19,7 +19,7 @@ search.addWidget(
 search.addWidget(
   instantsearch.widgets.refinementList({
     container: '#brand-list',
-    attributeName: 'brand',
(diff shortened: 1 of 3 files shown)


InstantSearch.js/refresh-cache-periodically/src/app.js
```diff
@@ -28,7 +28,7 @@ search.addWidget(
     container: '#hits',
     templates: {
       item: `
-        <div class="ais-Hits-item">
+        <div>
           <header class="hit-name">
             {{{_highlightResult.name.value}}}
```
samouss: We should migrate to the highlight function.

francoischalifour: Done in dd77325fa5b30e7c54bb3786440cf9a730d66992.

PR #30, opened 2018-12-17
feat(js): update refresh-cache-user-action example

francoischalifour:
Demo: https://codesandbox.io/s/github/algolia/doc-code-samples/tree/js-refresh-cache-user-action/InstantSearch.js/refresh-cache-user-action

### DIFF
--- a/InstantSearch.js/refresh-cache-user-action/src/app.js
+++ b/InstantSearch.js/refresh-cache-user-action/src/app.js
@@ -19,7 +19,7 @@ search.addWidget(
 search.addWidget(
   instantsearch.widgets.refinementList({
     container: '#brand-list',
-    attributeName: 'brand',
+    attribute: 'brand',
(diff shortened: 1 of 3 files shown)


InstantSearch.js/refresh-cache-user-action/src/app.js
```diff
@@ -28,7 +28,7 @@ search.addWidget(
     container: '#hits',
     templates: {
       item: `
-        <div class="ais-Hits-item">
+        <div>
           <header class="hit-name">
             {{{_highlightResult.name.value}}}
```
samouss: We should migrate to the highlight function.

francoischalifour: Done in fe9075f230d59b1d1a9e1977a4e983bbdaa77fe6.

PR #48, opened 2019-02-15
chore(test): add some basic tests

tkrugg:
This PR adds CI that tests every sample can install and build.
Later we can add integration testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,22 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/node:7.10
+
+    working_directory: ~/repo
+
+    steps:
+      - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,60 @@
...
+      - run: yarn install
+
+      - save_cache:
+          paths:
+            - node_modules
+          key: v1-dependencies-{{ checksum "package.json" }}
+~
+~
+~
+~
```
Haroenv: these lines are probably not on purpose

tkrugg: WIP 😄

samouss: We don't have the draft pull request options, that's weird.

PR #88, opened 2019-04-29
fix(angular): use search client

francoischalifour:
## Description

This updates all the Angular examples to use `searchClient` instead of `appId` and `apiKey`.

## Related

- algolia/angular-instantsearch#496


Angular InstantSearch/debounced-search-box/package.json
```diff
@@ -45,6 +45,7 @@
     "karma-jasmine-html-reporter": "0.2.2",
     "protractor": "6.0.0",
     "ts-node": "8.1.0",
-    "tslint": "5.16.0"
+    "tslint": "5.16.0",
+    "typescript": "2.7.2"
```
Haroenv: why such a low version of typescript?

Angular InstantSearch/geo-search/package.json
```diff
@@ -33,7 +33,6 @@
     "@angular/cli": "7.3.8",
     "@angular/compiler-cli": "6.1.10",
     "@angular/language-service": "6.1.10",
-    "@types/algoliasearch": "3.27.8",
```
Haroenv: why was this removed?

Angular InstantSearch/getting-started/src/app/app.component.ts
```diff
@@ -1,4 +1,10 @@
 import { Component } from '@angular/core';
+import algoliasearch from 'algoliasearch/lite';
```
Haroenv: inconsistent prettier formatting here, the other ones don't set single quote

Angular InstantSearch/manipulating-lists/filtering/src/app/app.component.ts
```diff
@@ -1,5 +1,11 @@
 import { Component } from '@angular/core';
 import { sortBy } from 'lodash';
+import algoliasearch from 'algoliasearch/lite';
```
Haroenv: same comment about prettier config

Angular InstantSearch/manipulating-lists/sorting/src/app/app.component.ts
```diff
@@ -1,5 +1,11 @@
 import { Component } from '@angular/core';
 import { sortBy } from 'lodash';
+import algoliasearch from 'algoliasearch/lite';
```
Haroenv: quotes

Angular InstantSearch/debounced-search-box/package.json
francoischalifour: I didn't change that. Yarn reordered the deps.

Angular InstantSearch/getting-started/src/app/app.component.ts
francoischalifour: It is consistent with the file. We need to run Prettier on another PR.

Angular InstantSearch/manipulating-lists/filtering/src/app/app.component.ts
francoischalifour: It is consistent with the file. We need to run Prettier on another PR.

Angular InstantSearch/manipulating-lists/sorting/src/app/app.component.ts
francoischalifour: It is consistent with the file. We need to run Prettier on another PR.

Angular InstantSearch/geo-search/package.json
francoischalifour: Because we import `algoliasearch/lite` which does not have the types. So it's unused.

Haroenv: lite does have the types though: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/algoliasearch/lite/index.d.ts

francoischalifour: Oh interesting! So I have this error in the latest version:

> ERROR in node_modules/@types/algoliasearch/lite/index.d.ts(14,43): error TS2304: Cannot find name 'Exclude'.

This is the error raised in algolia/algoliasearch-client-javascript#753.

We can move forward with this PR and add them when fixed.

Haroenv: As discussed, this is because the example uses 2.7, instead of a more recent version (2.8+) which comes with Exclude, which is written in the preamble of the definitions

francoischalifour:
Moving forward with this PR.

We can reiterate on the Angular examples by upgrading TypeScript to ≥ 2.8 and adding the development dependency `@types/algoliasearch`. See https://github.com/algolia/algoliasearch-client-javascript/issues/753#issuecomment-487562574 for context.

PR #102, opened 2019-05-13
feat(js): update SEO-friendly routing example

francoischalifour:
This updates the SEO-friendly routing InstantSearch.js example with a more appropriate URL schema. This example will be the base of the new [Routing URLs](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/js/) guide.

### Before

```sh
# Encoded
https://website.com/?menu%5Bcategories%5D=Cell%20Phones&refinementList%5Bbrand%5D%5B0%5D=Apple&refinementList%5Bbrand%5D%5B1%5D=Samsung&query=galaxy=Samsung&page=2&query=galaxy
# Decoded
https://website.com/?menu[categories]=Cell Phones&refinementList[brand][0]=Apple&refinementList[brand][1]=Samsung&page=2&query=galaxy
```

### After

```sh
# Encoded
https://website.com/Cell-Phones/?q=galaxy&p=2&brands=Apple%7CSamsung
# Decoded
https://website.com/Cell-Phones/?q=galaxy&p=2&brands=Apple|Samsung
```

## Preview

[Try it live →](https://codesandbox.io/s/yjojj4lnrz)


Haroenv:
We should wait to merge this until the guide is updated right?

InstantSearch.js/routing-seo-friendly/index.html
```diff
@@ -25,6 +25,10 @@
     <div class="container">
       <div class="search-panel">
         <div class="search-panel__filters">
+          <h3>Category</h3>
```
Haroenv: would be nice to do this with a panel instead no?

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,86 @@
...
+    };
+  },
+});
+
+const stateMapping = {
+  stateToRoute(uiState) {
+    return {
+      q: uiState.query || '',
+      p: uiState.page || 1,
+      category: (uiState.menu && uiState.menu.categories) || '',
```
Haroenv:
```suggestion
      category: (uiState.menu && uiState.menu.categories) || undefined,
```

Haroenv: otherwise you change the type

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,86 @@
...
+
+function getCategoryName(slug) {
+  return slug.replace(/-/g, ' ');
+}
+
+const router = instantsearch.routers.history({
+  windowTitle({ category, q }) {
+    const query = q ? `Results for "${q}"` : 'Search';
+
+    return [category, query].filter(Boolean).join(' – ');
```
Haroenv: while smart, I had to read it twice to understand what it was doing, maybe just an if would make sense?

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,86 @@
...
+    return {
+      query: routeState.q,
+      page: routeState.p,
+      menu: {
+        categories: routeState.category,
+      },
+      refinementList: {
+        brand:
+          (routeState.brands &&
+            routeState.brands !== 'all' &&
```
Haroenv: where does this `all` come from?

InstantSearch.js/routing-seo-friendly/src/search-router.js
francoischalifour: It makes more sense to join them here. We could have more parts in the title. Besides, an implementation with an `if` would be more difficult to understand IMO because the `–` dash shouldn't be displayed if one of the items are not defined.

Haroenv: I mean:

```js
if (category) {
  return  [category, query].join(' – ');
}
return query
```

francoischalifour:
@Haroenv Yes, we should synchronize the guide and this example's releases. This still needs to be fully reviewed before finalizing the new documentation (to avoid having multiple back and forths).

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,85 @@
...
+      addQueryPrefix: true,
+    });
+
+    return `${protocol}//${hostname}${portWithPrefix}${pathname}${queryString}${hash}`;
+  },
+  parseURL({ qsModule, location }) {
+    const routeStateString = location.pathname.split('/')[1];
+    const category = getCategoryName(routeStateString.split('/')[0]);
+    const queryParameters = qsModule.parse(location.search.slice(1));
+    const { q = '', p, brands } = queryParameters;
```
samouss: Is there a reason why we don’t assign on those values on the line above?

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,85 @@
...
+    const { protocol, hostname, port = '', hash } = location;
+    const portWithPrefix = port === '' ? '' : `:${port}`;
+    const pathname = routeState.category
+      ? `/${getCategorySlug(routeState.category)}/`
+      : '';
+
+    const searchState = {};
+    if (routeState.q) searchState.q = encodeURIComponent(routeState.q);
+    if (routeState.p !== 1) searchState.p = routeState.p;
+    if (routeState.brands) searchState.brands = routeState.brands;
```
samouss: Is there a reason why we don't encode the brand values? They might also contain space or weird character.

InstantSearch.js/routing-seo-friendly/src/search-router.js
francoischalifour: Mainly for clarity for our users.

InstantSearch.js/routing-seo-friendly/src/search-router.js
francoischalifour: Nope, I added that.

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,85 @@
...
+    if (category) {
+      return [category, query].join(' – ');
+    }
+
+    return query;
+  },
+  createURL({ qsModule, routeState, location }) {
+    const { protocol, hostname, port = '', hash } = location;
+    const portWithPrefix = port === '' ? '' : `:${port}`;
+    const pathname = routeState.category
```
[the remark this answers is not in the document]
mthuret: Not sure to get the full context here/questions. cc @francoischalifour @samouss

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -24,13 +24,19 @@ const router = instantsearch.routers.history({
...
-    if (routeState.p !== 1) searchState.p = routeState.p;
-    if (routeState.brands) searchState.brands = routeState.brands;
+    if (routeState.q) {
+      queryParameters.q = encodeURIComponent(routeState.q);
+    }
+    if (routeState.p !== 1) {
+      queryParameters.p = routeState.p;
+    }
+    if (routeState.brands) {
+      queryParameters.brands = encodeURIComponent(routeState.brands);
```
samouss: You have to decode it otherwise the sync is not working anymore. Note that we also encode the delimiter `|` not sure it's the expected behaviour. We might want to only encode the value but not the delimiter.

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,85 @@
+/* global instantsearch */
+
+function getCategorySlug(name) {
+  return name.replace(/\s/g, '-');
+}
+
+function getCategoryName(slug) {
+  return slug.replace(/-/g, ' ');
```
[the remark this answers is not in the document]
tkrugg: Isn't it more common to replace spaces with `+` ?
Quoting marie's example:
https://www.galerieslafayette.com/c/maison-cuisine+et+arts+de+la+table-arts+de+la+table?s=vaisselle

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,91 @@
...
+    return {
+      q: decodeURIComponent(q),
+      p,
+      category: decodeURIComponent(category),
+      brands,
+    };
+  },
+});
+
+const stateMapping = {
```
tkrugg: This is still a lot of code and some of it isn't always easy to read.
I would suggest maybe we reduce the complexity of the example by 
- removing `q` for `query` and just keeping `query` everywhere
- removing `p` for `page` and just keeping `page` everywhere
- removing `brands` or any kind of refinement list
     * because I don't see people adding theses kind of links to their sitemaps  `mywebsite.com/phones/?brands=Apple|Samsung`, but I may be wrong
     * it removes this chunky part `uiState.refinementList && uiState.refinementList.brand && uiState.refinementList.brand.join('|')`

InstantSearch.js/routing-seo-friendly/src/search-router.js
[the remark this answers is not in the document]
samouss: I've got the approbation of @mthuret on this one. We should support path like `/search`.

InstantSearch.js/routing-seo-friendly/src/search-router.js
francoischalifour: We're now repeating the array values:

```
?brands=Metra&brands=Samsung
```

InstantSearch.js/routing-seo-friendly/src/search-router.js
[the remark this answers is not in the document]
francoischalifour: I added support for that.

InstantSearch.js/routing-seo-friendly/src/search-router.js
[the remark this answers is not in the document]
francoischalifour: Updated!

InstantSearch.js/routing-seo-friendly/src/search-router.js
francoischalifour: Agreed on Slack to go for:

```
https://website.com/search/Cell+Phones/?query=m&page=4&brands=Metra&brands=Samsung
```

Let me know what you think.

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,101 @@
...
+
+    return `${baseUrl}search/${categoryPath}${queryString}`;
+  },
+
+  parseURL({ qsModule, location }) {
+    const pathnameMatches = location.pathname.match(/search\/(.*?)\/?$/);
+    const category = getCategoryName(
+      (pathnameMatches && pathnameMatches[1]) || ''
+    );
+    const { query = '', page, brands = [] } = qsModule.parse(
```
Haroenv: how come you don't pass the same options as to the stringify?

francoischalifour: What do you mean?

Haroenv: you don't pass arrayMerge here, is it taken automatically?

francoischalifour: Where do you see such an option in `qs`?

That's how it's transformed back into an array if needed:

https://github.com/algolia/doc-code-samples/blob/0d1464f5b9af7997ebf9975e5c702044c916d1a9/InstantSearch.js/routing-seo-friendly/src/search-router.js#L57-L61

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -50,38 +47,35 @@ const router = instantsearch.routers.history({
...
+    const { query = '', page, brands } = queryParameters;
 
     return {
-      q: decodeURIComponent(q),
-      p,
+      query: decodeURIComponent(query),
+      page,
       category: decodeURIComponent(category),
-      brands: decodeURIComponent(brands),
+      brands: brands && brands.map(decodeURIComponent),
```
samouss: We could use `[]` otherwise an empty value is selected when brands is without value.

> https://yjojj4lnrz.codesandbox.io/search?query=apple&brands

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,101 @@
+/* global instantsearch */
+
+function getCategorySlug(name) {
+  return name.replace(/\s/g, '+');
```
samouss: We only "slugify" the spaces but what about punctation? It's maybe safer to rely on a third party.

> https://yjojj4lnrz.codesandbox.io/search/Ranges,+Cooktops+&+Ovens/?query=apple

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,101 @@
...
+
+    if (category) {
+      return `${category} – ${queryTitle}`;
+    }
+
+    return queryTitle;
+  },
+
+  createURL({ qsModule, routeState, location }) {
+    let baseUrl = location.href.split('/search/')[0];
```
samouss: I don't know how to reproduce it but I manage to have an URL with multiple times `/search` in it.

InstantSearch.js/routing-seo-friendly/src/search-router.js
Haroenv: I think the problem for those is that if you remove multiple things by the same, the conversion is lossy. Those issues should be mostly solved by encodeURIComponent I think

InstantSearch.js/routing-seo-friendly/src/search-router.js
Haroenv: https://yjojj4lnrz.codesandbox.io/search?query=apple&brands/search/?query=search&brands=

if the query is "search"

InstantSearch.js/routing-seo-friendly/src/search-router.js
francoischalifour: We could still encode/decode the category instead of manually do it. It will however be less readable.

What do we value the most?

InstantSearch.js/routing-seo-friendly/src/search-router.js
Haroenv: (ignore this, I can't find how I got this url anymore) :'(

InstantSearch.js/routing-seo-friendly/src/search-router.js
samouss: We can check on other websites how it's implemented. The safe bet is to use the encoded version. But we can mix both solutions: apply `encodeURIComponent` to each part of the value then apply the slug. For most cases, the encoding will be a noop which means that we have a readable URL.

francoischalifour: In 507ee489e233c8365e3b70168da782e26b96f034 I slugified the category and encoded/decoded.

InstantSearch.js/routing-seo-friendly/src/app.js
```diff
@@ -1,85 +1,15 @@
 /* global instantsearch algoliasearch */
+import searchRouter from './search-router';
```
samouss: What about `searchRouting` rather than `searchRouter` ?

It is is a bit confusing because one of the option available for `routing` is called `router`.

InstantSearch.js/routing-seo-friendly/src/search-router.js
samouss: The issue occurs when we omit the trailing slash inside the URL. Click on the URL below and trigger a refinement, the URL should not contain two times `/search`. The last occurence is the correct URL.  We could use a RegExp to avoid this issue.

> https://yjojj4lnrz.codesandbox.io/search?query=apple

InstantSearch.js/routing-seo-friendly/src/search-router.js
```diff
@@ -0,0 +1,135 @@
...
+      location.search.slice(1)
+    );
+    // `qs` does not return an array when there's a single value.
+    const allBrands = Array.isArray(brands) ? brands : [brands].filter(Boolean);
+
+    return {
+      query: decodeURIComponent(query),
+      page,
+      brands: allBrands.map(decodeURIComponent),
+      category: decodeURIComponent(category),
```
samouss: The `category` is already decoded inside the function `getCategoryName`.

InstantSearch.js/routing-seo-friendly/src/app.js
francoischalifour: Renamed!

InstantSearch.js/routing-seo-friendly/src/search-router.js
francoischalifour: Fixed in 21788d37a6a9f6f1a90c72cd0cbe17bb389a28f3.

PR #104, opened 2019-05-16
chore(Angular InstantSearch): use InstantSearch + Insights integration in algolia-insights sample

tkrugg:
Other changes are:
- update search-insights to v1.2.0
- update search-insights install snippets to extract source URL (as
documented here: https://github.com/algolia/doc/pull/3063)


Angular InstantSearch/algolia-insights/src/app/app.component.ts
```diff
@@ -16,13 +16,7 @@ const searchClient = algoliasearch(
 export class AppComponent {
   config = {
     indexName: "demo_ecommerce",
-    searchClient
+    searchClient,
+    insightsClient: aa
```
samouss:
```suggestion
    insightsClient: window.aa
```

Unless TypeScript complains?

tkrugg: typescript doesn't like it. It returns `Property 'aa' does not exist on type 'Window'.`
I can use `(window as any).aa` however.

tkrugg: done here: fa014f0

PR #105, opened 2019-05-17
chore(React InstantSearch): use InstantSearch + Insights integration in algolia-insights sample

tkrugg:
Other changes are:
- remove custom Hits component as no longer necessary
- update search-insights to v1.2.0
- update search-insights install snippets to extract source URL (as documented here: https://github.com/algolia/doc/pull/3063)


React InstantSearch/algolia-insights/src/App.js
```diff
@@ -20,6 +20,8 @@ const searchClient = algoliasearch(
   'aadef574be1f9252bb48d4ea09b5cfe5'
 );
 
+const HitWithInsights = connectHitInsights(window.aa)(Hit);
```
Haroenv: I'd move this closer to (probably just after) `Hit`, right now this order confused me

React InstantSearch/algolia-insights/src/App.js
```diff
@@ -71,14 +59,8 @@ function Hit({ hit, index, searchResults }) {
       <button
         className="hit-action"
         onClick={() => {
-          window.aa('clickedObjectIDsAfterSearch', {
-            index: 'demo_ecommerce',
+          insights('clickedObjectIDsAfterSearch', {
             eventName: 'Add to favorite',
-            queryID: searchResults.queryID,
```
Haroenv: <3 these changes!

React InstantSearch/algolia-insights/src/App.js
samouss: I agree @Haroenv, it's probably better to move the `Hit` inside its own file.

tkrugg: done here: 41da299

PR #111, opened 2019-05-22
fix: typo

tkrugg:


Vue InstantSearch/server-side-rendering/src/entry-server.js
```diff
@@ -23,8 +23,7 @@ export default context => {
           context.algoliaState = instantsearch.getState();
 
           resolve(app);
```
Haroenv: would maybe make more sense if it was removed here, to stay with the Vue CLI example

tkrugg: sure thing.

PR #128, opened 2019-06-13
feat(react): update SEO-friendly routing example

francoischalifour:
This updates the SEO-friendly routing React InstantSearch example with a more appropriate URL schema. This example will be the base of the new [Routing URLs](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/react/) guide.

### Before

```sh
https://website.com/?menu[categories]=Cell Phones&refinementList[brand][0]=Apple&refinementList[brand][1]=Samsung&page=2&query=galaxy
```

### After

```sh
https://website.com/Cell+Phones/?query=galaxy&page=2&brands=Apple&brands=Samsung
```

## Preview

[See preview →](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/react-seo-routing/React+InstantSearch/routing-seo-friendly)

### DIFF
--- a/React InstantSearch/routing-seo-friendly/src/App.js
+++ b/React InstantSearch/routing-seo-friendly/src/App.js
@@ -67,8 +67,8 @@ const createURL = state => {
   const isDefaultRoute =
     !state.query &&
     state.page === 1 &&
-    state.refinementList.brand.length === 0 &&
-    !state.menu.categories;
+    (state.refinementList && state.refinementList.brand.length === 0) &&
+    (state.menu && !state.menu.categories);
 
   if (isDefaultRoute) {
     return '';
--- a/React InstantSearch/routing-seo-friendly/src/App.js
+++ b/React InstantSearch/routing-seo-friendly/src/App.js
@@ -1,11 +1,14 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import {
   InstantSearch,
   Hits,
   SearchBox,
+  Menu,
   RefinementList,
   Pagination,
   Highlight,
+  Panel,
+  ClearRefinements,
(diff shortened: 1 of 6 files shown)


React InstantSearch/routing-seo-friendly/src/App.js
```diff
@@ -1,11 +1,14 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
```
samouss: We do use the hooks inside storybook but we've never used them for documentation purpose. I'm wondering if it's the right time to do the changes. It's a different paradigm that might be complicated to understand at first from our users. I would have wait a bit more for their introduction into the doc.

React InstantSearch/routing-seo-friendly/src/App.js
```diff
@@ -18,118 +21,171 @@ const searchClient = algoliasearch(
...
+    .split('+')
+    .map(decodeURIComponent)
+    .join(' ');
+}
+
+const createURL = state => {
   const isDefaultRoute =
     !state.query &&
     state.page === 1 &&
     state.refinementList.brand &&
```
samouss: `state.refinementList` is always defined?

francoischalifour: Yes, actually we can even simplify the condition to:

```js
const isDefaultRoute =
  !state.query &&
  state.page === 1 &&
  state.refinementList.brand.length === 0 &&
  !state.menu.categories;
```

samouss: It's always defined in our case because the `RefinementList` is always mounted. Since it's an example I would over check because end users might have UIs that does not always have those widgets.

PR #132, opened 2019-06-28
feat(angular): update SEO-friendly routing example

tkrugg:
Replicated changed documented for the vanilla PR https://github.com/algolia/doc-code-samples/pull/102.


Angular InstantSearch/routing-seo-friendly/src/app/search/search.component.ts
```diff
@@ -0,0 +1,125 @@
...
+
+          if (category) {
+            return `${category} – ${queryTitle}`;
+          }
+
+          return queryTitle;
+        },
+
+        createURL({ qsModule, routeState, location }) {
+          const urlParts = location.href.match(/^(.*?)\/search/);
```
samouss: We can't leverage the router of Angular for those operations?

tkrugg: done here: 0d935f8

Angular InstantSearch/routing-seo-friendly/src/app/search/search.component.ts
```diff
@@ -0,0 +1,120 @@
...
+          if (routeState.brands) {
+            queryParameters.brands = routeState.brands.map(encodeURIComponent);
+          }
+
+          const queryString = qsModule.stringify(queryParameters, {
+            addQueryPrefix: true,
+            arrayFormat: "repeat"
+          });
+
+          return `/search/${categoryPath}${queryString}`;
```
francoischalifour: This likely won't work with a custom pathname.

E.g.: `https://angular-instantsearch.netlify.com/examples/e-commerce/`

It will override `examples/e-commerce`.

We need to prepend the original pathname. Something like this:

```js
const urlParts = location.pathname.match(/^(.*?)\/search/);
const originalPathName = urlParts ? urlParts[1] : '';
```

tkrugg: I understand we need that for the demos we published on `https://community.algolia.com/angular-instantsearch/examples/*` but that's rather a details related to the way we publish our examples.

This example runs on `http://localhost:<port>/search` and this implementation of createURL simple and completely sufficient.

francoischalifour: Go ahead then!

FYI you can see the issue when deployed on algolia/instantsearch.js#3921.

PR #134, opened 2019-07-12
fix: use router-link

samouss:
This PR uses the `routerLink` directive rather than `href`. 

It prevents the browser to completely refresh the page when we transition from one page to another.


Haroenv:
https://codesandbox.io/s/github/algolia/doc-code-samples/tree/fix/router-link/Angular%20InstantSearch/routing-angular-router

vs old https://codesandbox.io/s/github/algolia/doc-code-samples/tree/master/Angular%20InstantSearch/routing-angular-router

PR #136, opened 2019-07-12
chore: update TypeScript to 2.8.4

samouss:
This PR bumps TypeScript to `2.8.4`. It allows us to use the latest `@types/algoliasearch`.


samouss:
Because I didn't want to break other possible packages.

PR #141, opened 2019-07-23
feat(react): add multi-index-autocomplete-with-insights

tkrugg:
⚠️ this is waiting for this to be merged: https://github.com/algolia/react-instantsearch/pull/2687/files

Copied multi-index-autocomplete and added insights to it.


Haroenv:
if we write a guide on this, it can be recreated, but for the time being, it's probably better off being closed.

PR #152, opened 2020-01-30
update Vue InstantSearch for v3

Haroenv:
- [x] multi index hits
- [x] multi index autocomplete
- [x] query suggestions
- [x] update dependencies
- [x] routing-full-url
- [x] routing-basic
- [x] routing-vue-router
- [x] routing-state-mapping

This PR is blocked until the library is released & the docs are live

### DIFF
--- a/Vue InstantSearch/autocomplete-results-page/src/App.vue
+++ b/Vue InstantSearch/autocomplete-results-page/src/App.vue
@@ -15,38 +15,35 @@
         :search-client="searchClient"
         index-name="demo_ecommerce"
       >
-        <ais-configure
-          :hitsPerPage="5"
-          :restrictSearchableAttributes="['name']"
-        />
-        <ais-autocomplete>
-          <template slot-scope="{ currentRefinement, indices, refine }">
-            <vue-autosuggest
-              :suggestions="indicesToSuggestions(indices)"
-              @selected="onSelect"
-              :input-props="{
-                style: 'width: 100%',
-                onInputChange: refine,
-                placeholder: 'Search here…',
-              }"
-            >
-              <template slot-scope="{ suggestion }">
-                <ais-highlight
-                  :hit="suggestion.item"
-                  attribute="name"
-                  v-if="suggestion.item.name"
-                />
-                <strong>$ {{ suggestion.item.price }}</strong>
-                <img :src="suggestion.item.image" />
-              </template>
-            </vue-autosuggest>
-          </template>
-        </ais-autocomplete>
-      </ais-instant-search>
-      <ais-instant-search
-        :search-client="searchClient"
-        index-name="demo_ecommerce"
-      >
+        <ais-index index-name="demo_ecommerce">
+          <ais-configure
+            :hitsPerPage="5"
+            :restrictSearchableAttributes="['name']"
+          />
+          <ais-autocomplete>
+            <template slot-scope="{ currentRefinement, indices, refine }">
+              <vue-autosuggest
+                :suggestions="indicesToSuggestions(indices)"
+                @selected="onSelect"
+                :input-props="{
+                  style: 'width: 100%',
+                  onInputChange: refine,
+                  placeholder: 'Search here…',
+                }"
+              >
+                <template slot-scope="{ suggestion }">
+                  <ais-highlight
+                    :hit="suggestion.item"
+                    attribute="name"
+                    v-if="suggestion.item.name"
+                  />
+                  <strong>$ {{ suggestion.item.price }}</strong>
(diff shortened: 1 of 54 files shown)


Vue InstantSearch/autocomplete-results-page/src/App.vue
```diff
@@ -15,38 +15,35 @@
...
-              </template>
-            </vue-autosuggest>
-          </template>
-        </ais-autocomplete>
-      </ais-instant-search>
-      <ais-instant-search
-        :search-client="searchClient"
-        index-name="demo_ecommerce"
-      >
+        <ais-index index-name="demo_ecommerce">
```
Haroenv: this method now always does two queries, of which one unchanged, since the autocomplete index doesn't use the parent query. Do we leave this, or revert to the old two-instance method?

Vue InstantSearch/query-suggestions/src/App.vue
```diff
@@ -13,55 +13,56 @@
     <div class="container">
       <ais-instant-search
         :search-client="searchClient"
-        index-name="instant_search_demo_query_suggestions"
+        index-name="instant_search"
```
Haroenv: this does a useless query at root, since the autocomplete index should _not_ have the disjunctive facets set by the root / now sibling.

Vue InstantSearch/query-suggestions/src/App.vue
```diff
@@ -13,55 +13,56 @@
...
-          <template slot-scope="{ currentRefinement, indices, refine }">
-            <vue-autosuggest
-              :suggestions="indicesToSuggestions(indices)"
-              @selected="onSelect"
-              :input-props="{
-                style: 'width: 100%',
-                onInputChange: refine,
-                placeholder: 'Search here…',
-              }"
+        <ais-index index-name="instant_search_demo_query_suggestions">
```
Haroenv: having both indices means that both will always be queried, even though only one at a time changes.

Vue InstantSearch/query-suggestions/src/App.vue
Haroenv: I could filter the main index out with a custom search client, but not sure that's the smartest idea to put in a guide

Vue InstantSearch/query-suggestions/src/App.vue
```diff
@@ -95,20 +96,25 @@ export default {
...
-            ...hit,
-            category,
-            name: hit.query,
-          };
-        }),
-      }));
+      return indices
+        .filter(
+          ({ indexName }) =>
+            indexName === 'instant_search_demo_query_suggestions'
```
Haroenv: since the other index is now adjacent, you also get access to its results in the autocomplete index.

Haroenv:
to run these, you need to do

```sh
yarn add @haroenv/not-vue-instant-search
```

and change the import in `src/main.js`

eunjae-lee:
Is it a privately published package?

```
➜ yarn add @haroenv/not-vue-instant-search
yarn add v1.16.0
[1/4] 🔍  Resolving packages...
error An unexpected error occurred: "https://registry.npmjs.org/@haroenv%2fnot-vue-instant-search: Not found".
info If you think this is a bug, please open a bug report with the information provided in "/Users/eunjaelee/workspace/doc-code-samples/Vue InstantSearch/autocomplete-results-page/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.


➜ npm install @haroenv/not-vue-instant-search
npm ERR! code E404
npm ERR! 404 Not Found: @haroenv/not-vue-instant-search@latest

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/eunjaelee/.npm/_logs/2020-02-05T12_43_16_678Z-debug.log
```

Haroenv:
I meant `@haroenv/not-vue-instantsearch`

Vue InstantSearch/autocomplete-results-page/src/App.vue
eunjae-lee: Could you help understand it a little?
Kapture 2020-02-05 at 14 57 52
When I tried, each stroke triggered only one query at a time. And when I removed the wrapping `ais-index`, it was the same and worked fine.

So my question is

1. What do you mean by two queries?
2. Why do we need `ais-index` component in this example?

Haroenv: a single request, with two queries in it

Haroenv: We don't _need_ ais-index, but it should be an example of not needing two search instances.

PR #232, opened 2020-10-20
feat(autocomplete): add autocomplete example with recent searches

eunjae-lee:
## Summary

This PR adds an example for "Autocomplete with recent searches".
It has been updated with the recent release of Autocomplete.js.

[[Preview →](https://codesandbox.io/s/github/algolia/doc-code-samples/blob/feat/plugins/Autocomplete/multi-source)]


Autocomplete/predefined-items-and-recent-searches/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+              {
+                indexName: "instant_search_demo_query_suggestions",
+                query,
+                params: {
+                  hitsPerPage: 4,
+                  facetFilters: [...recentSearches.data.getFacetFilters()],
+                },
+              },
+            ],
+          }).then((results) => {
```
eunjae-lee, line 42: We can get rid of this once https://github.com/algolia/autocomplete.js/pull/331 is released.

Autocomplete/predefined-items-and-recent-searches/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+import { recentSearches } from "./recent-searches";
+
+const searchClient = algoliasearch(
+  "latency",
+  "6be0576ff61c053d5f9a3225e2a90f76"
+);
+
+autocomplete({
+  container: "#autocomplete",
+  plugins: [predefinedItems, recentSearches],
```
eunjae-lee, line 17: For now, the order is always `plugins` -> `getSources` here.
In order to display the predefined items before recent searches, I created a plugin for the predefined items.
Does it make sense to you?

Autocomplete/predefined-items-and-recent-searches/src/predefined-items.js
```diff
@@ -0,0 +1,51 @@
+export const predefinedItems = {
+  getSources({ query }) {
+    if (query) {
```
eunjae-lee, line 3: If user starts typing something, this part disappears.

Autocomplete/predefined-items-and-recent-searches/src/recent-searches.js
```diff
@@ -0,0 +1,29 @@
+import { createRecentSearchesPlugin } from '@algolia/autocomplete-plugin-recent-searches';
+const orgRecentSearches = createRecentSearchesPlugin({
+  key: 'RECENT_SEARCH',
+  limit: 5,
+});
+
+export const recentSearches = {
```
eunjae-lee, line 7: Currently `@algolia/autocomplete-plugin-recent-searches` doesn't allow custom template manipulation. That's why I'm spreading and injecting the header template.
Should we introduce an API in autocomplete to allow altering templates for plugins?

Autocomplete/predefined-items-and-recent-searches/src/style.css
```diff
@@ -0,0 +1,168 @@
+* {
```
eunjae-lee, line 1: Do we want to provide a default css file for autocomplete-js? cc @francoischalifour

Autocomplete/predefined-items-and-recent-searches/src/index.js
Haroenv: I think it could be fixed if a plugin would expose sources instead. Is there a reason we didn't go with that design?

eunjae-lee: If you look at [this custom plugin](Autocomplete/predefined-items-and-recent-searches/src/predefined-items.js), a plugin is actually an object with `getSources` function. If you put like in `plugins` array, the autocomplete-core takes care of calling the sources, and some other lifecycle hooks.
It means if user wants, they can do things like:

```js
autocopmlete({
  ...
   getSources(params) {
    return recentSearches.getSources(params);
  }
})
```

Autocomplete/predefined-items-and-recent-searches/src/style.css
francoischalifour: Yes—we starting working on a theme with Kevin before the whole Documentation Search v2. Hopefully we'll starting working back on it soon.

Autocomplete/predefined-items-and-recent-searches/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+const searchClient = algoliasearch(
+  'latency',
+  '6be0576ff61c053d5f9a3225e2a90f76'
+);
+
+autocomplete({
+  container: '#autocomplete',
+  plugins: [predefinedItems, recentSearches],
+  openOnFocus: true,
+  onStateChange({ state, prevState }) {
```
francoischalifour, line 19: Is that supposed to be here?

Autocomplete/predefined-items-and-recent-searches/src/predefined-items.js
```diff
@@ -0,0 +1,51 @@
...
+          header() {
+            return `
+              <div class="aa-PredefinedHeader">
+                <p>Useful Links</p>
+              </div>
+            `;
+          },
+          item({ item, root }) {
+            return `
+              <div class="aa-PredefinedItem">
```
francoischalifour, line 38: We should make the whole container a link (wrapping the `div` with `a`).

Autocomplete/predefined-items-and-recent-searches/src/recent-searches.js
```diff
@@ -0,0 +1,29 @@
+import { createRecentSearchesPlugin } from '@algolia/autocomplete-plugin-recent-searches';
+const orgRecentSearches = createRecentSearchesPlugin({
```
francoischalifour, line 2: `org` confused me at first—could stand for "organization". `original` would be more explicit.

Autocomplete/predefined-items-and-recent-searches/src/recent-searches.js
francoischalifour: Yes—we need to find a proper solution for this.

Autocomplete/predefined-items-and-recent-searches/src/predefined-items.js
```diff
@@ -0,0 +1,51 @@
+export const predefinedItems = {
```
francoischalifour, line 1: That's interesting to treat this as a plugin 🙂

Autocomplete/predefined-items-and-recent-searches/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+        getSuggestions({ query }) {
+          return getAlgoliaHits({
+            searchClient,
+            queries: [
+              {
+                indexName: 'instant_search_demo_query_suggestions',
+                query,
+                params: {
+                  hitsPerPage: 4,
+                  facetFilters: [...recentSearches.data.getFacetFilters()],
```
Meschreiber, line 37: I'm not sure what the need for these facetFilters is?

Autocomplete/predefined-items-and-recent-searches/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+            return results[0];
+          });
+        },
+        templates: {
+          header() {
+            if (query) {
+              return '';
+            }
+            return `
+              <div class="aa-HitsHeader">
```
Meschreiber, line 51: Do you think it makes sense to remove the “Search Results” header added for the initial results shown? These are really no different than the suggestions made once the user starts typing, so I think it might be confusing to have the header in on place, but not the other.

Autocomplete/predefined-items-and-recent-searches/src/index.js
eunjae-lee: It's to demonstrate for users a way to get notified of query change. This example isn't a standalone, but they need to do something else with it.

Autocomplete/predefined-items-and-recent-searches/src/predefined-items.js
eunjae-lee: Can you tell me why?

Autocomplete/predefined-items-and-recent-searches/src/predefined-items.js
eunjae-lee: yeah, everything can be a plugin :)

Autocomplete/predefined-items-and-recent-searches/src/index.js
eunjae-lee: So this is to exclude the items from recent searches in the actual hits.

```js
recentSearches.data.getFacetFilters()
```

returns an array like
```json
[
  'objectID:-aaaaa',
  'objectID:-bbbbb',
  'objectID:-ccccc',
  ...
]
```

Autocomplete/predefined-items-and-recent-searches/src/predefined-items.js
francoischalifour: Otherwise if you click anywhere after the "GitHub repository" text, the dropdown closes without redirecting to the link.

image

Autocomplete/predefined-items-and-recent-searches/src/predefined-items.js
```diff
@@ -0,0 +1,51 @@
...
+  getSources({ query }) {
+    if (query) {
+      return [];
+    }
+
+    return [
+      {
+        getSuggestions() {
+          return [
+            {
```
francoischalifour, line 11: If we want to truly make this a reusable plugin, should this array be passed as an option of the plugin?

Autocomplete/predefined-items-and-recent-searches/src/index.js
Meschreiber: Ah, gtk, thanks! Will add as a comment in https://github.com/algolia/doc/pull/5033

Autocomplete/predefined-items-and-recent-searches/src/predefined-items.js
eunjae-lee: Yes, but this is an example project. So I didn't go there.

eunjae-lee:
Hey @francoischalifour @Haroenv @Meschreiber ,
I've updated this branch to apply the latest changes in Autocomplete.js.

This branch used to be an example for autocomplete + predefined items + recent searches, 
but now "predefined items" is no longer here.

Once this PR is merged, I will copy this code to come up with an example for predefined items.

Autocomplete/recent-searches/src/index.js
```diff
@@ -0,0 +1,39 @@
...
+  '6be0576ff61c053d5f9a3225e2a90f76'
+);
+const index = searchClient.initIndex('instant_search');
+
+autocomplete({
+  container: '#autocomplete',
+  plugins: [recentSearches],
+  openOnFocus: true,
+  onSubmit({ state }) {
+    index.search(state.query).then((result) => {
```
Haroenv, line 21: I feel like the results should also be shown on any input, typing in an input and it doing nothing was very confusing for me at first. Maybe you can put it in a function like `searchAndShowMainResults` to make it easier to replace the function call with an InstantSearch refine

eunjae-lee: Non search-as-you-type was kind of confusing to me. You can try the [preview](https://codesandbox.io/s/github/algolia/doc-code-samples/blob/feat/plugins/Autocomplete/recent-searches) again. How is it now?

Autocomplete/recent-searches/src/index.js
```diff
@@ -0,0 +1,45 @@
...
+  });
+}
+
+autocomplete({
+  container: '#autocomplete',
+  plugins: [recentSearches],
+  openOnFocus: true,
+  onStateChange({ state, prevState }) {
+    if (state.query !== prevState.query) {
+      searchAndShowMainResults(state.query);
```
francoischalifour, line 42: We could better separate concerns:

```ts
const hitsContainer = document.querySelector('#hits');
// ...
index.search(query).then(result => render({ hits: result.hits }, hitsContainer));
```

francoischalifour (commented):
It feels weird to display only Recent Searches, we should probably also add the Query Suggestions plugin.

Autocomplete/recent-searches/src/index.js
Haroenv: maybe more clean like this with a note? https://codesandbox.io/s/lucid-torvalds-gwf3m?file=/index.html

eunjae-lee:
@francoischalifour 

> It feels weird to display only Recent Searches, we should probably also add the Query Suggestions plugin.

Yeah it's true, but this example is going to be used to explain how to implement recent searches. There will be another example with predefined items + recent searches + QS + search results.

Autocomplete/recent-searches/src/index.js
eunjae-lee: https://github.com/algolia/doc-code-samples/pull/232/commits/3f2db5658e62aa4c122db953937ed99112b651f2

Autocomplete/recent-searches/src/index.js
eunjae-lee: https://github.com/algolia/doc-code-samples/pull/232/commits/f2464cc0ef0ea99fa2b5dbd76afe22d0d34ed619

Haroenv:
Alternatively we could also add a button to fill the localstorage with some default values so users can see how it works? just recent searches alone seems a bit confusing

francoischalifour:
I've thought about alternatives and the version with QS is the one that I see would make the most sense for the users. Perhaps adding an "instruction" section:

- Search in the input
- Select a Query Suggestion
- Notice that this query is now stored in your recent searches

Besides, I would expect to have learned using Query Suggestions before following the Recent Searches guide.

Haroenv:
maybe even static results instead of query suggestions?

eunjae-lee:
Okay then, let ping @Meschreiber since she knows the best about the flow of the docs and whatnot.

francoischalifour:
> maybe even static results instead of query suggestions?

@Haroenv Are you suggesting this to make the code sample easier? [QS is an Autocomplete plugin](https://autocomplete.algolia.com/docs/createQuerySuggestionsPlugin) so there won't be much code to add FYI.

Meschreiber:
Thanks for the ping! Indeed, what we discussed with Solutions/GTM was trying to have as atomic as possible examples in the UI/UX Patterns section. These can then be composed into more complex Solutions. (cc @seafoox @codeharmonics @eunice  in case I'm misrepresenting that.) Based on the [convo in this PR](https://github.com/algolia/doc/pull/5006#discussion_r516104904), we agree to have these pages under UI/UX Patterns > Autocomplete:

- Autocomplete standalone - both single index and multi-index (Will subsume rich horizontal menu)
- Autocomplete + IS search - both single index and multi-index (There's a lot of overlap here with QS page --> still need to think more about this)
- Page each for search pattern (could be w/ InstantSearch or w/o)
    - Query Suggestions
    - Recent searches
    - Predefined/static searches

That said, I agree it's strange to have a demo with *just* recent searches; when a user first tries to sample the demo they won't see anything and be confused about it.

Maybe what we can do for each of the "search patterns" is show a code sample that integrates all three, but then highlight/focus the tutorial on just the particular plug-in we are describing.  That way the user has the benefit of seeing the the solution in context, but the tutorial is still bite-sized and focused on one thing.

The other alternative is to just have one page for all of the "search patterns" and call it plug-ins. If I understand correctly, they're all plug-ins and integrating them is pretty much the same thing/very similar, so it might be repetitive to have a separate page for each one. This goes against what we agreed with Solutions though, so I would prefer keeping them atomic if we can do it. (Feel free to weigh in here @seafoox @codeharmonics @eunice  )

eunjae-lee:
@Meschreiber I agree. Everything can be plugins in autocomplete v1, so we can have one or two complete example and in the doc, we can break it down to bits plugin by plugin and explain each plugin in its own context.

I could come up with an example of autocomplete + predefined items + recent searches + query suggestions. What do you think?

Meschreiber:
Sounds good to me @eunjae-lee ! To confirm, we then would use this (autocomplete + predefined items + recent searches + query suggestions) code sandbox for discrete guides on 1) predefined items 2) recent searches. (There will also be a separate page about autocomplete + QS, as far as I understand what we said in our discussions.)

eunjae-lee:
@Meschreiber Let me know what you think: https://codesandbox.io/s/github/algolia/doc-code-samples/blob/feat/plugins/Autocomplete/basic
cc @francoischalifour

Meschreiber:
Argh, sorry to miss this notif yesterday. This looks good @eunjae-lee!

Just a couple of things:
1) Magnifier/search button is covering the first character(s) of the input
2) I don't think we need to overlay the autocomplete on an InstantSearch page

I see I wrote this description from our meeting notes:
> Codesandbox for 3. -- queries QS index and products index and uses plug-ins for recent searches + static searches

When writing "queries products index", I was thinking it is querying the products index to display products in the autocomplete, but I think it's actually unnecessary to have that or the overlay on an IS results page. If we really want, we can have enter display products, but as is I think there may be too much going on, esp. for a "basic standalone" autocomplete

WDYT?

eunjae-lee:
@Meschreiber 

> 1. Magnifier/search button is covering the first character(s) of the input

It's fixed :)

> 2. I don't think we need to overlay the autocomplete on an InstantSearch page

Before answering your question, this example didn't use InstantSearch. It's autocomplete only, but it searches and display the hits in a "search-as-you-type" way. I think that's why you thought there's InstantSearch in this example.

```js
  onStateChange({ state, prevState }) {
    if (state.query !== prevState.query) {
      searchAndRenderResult(state.query);
    }
  },
```

When the state changes in autocomplete, it searches using searchClient and renders the hit by `container.innerHTML = ...`.

Meschreiber:
Ah yes, sorry, I totally used that term incorrectly. 😬  I often just think of an updating results page as "InstantSearch" even though that's not what this is, and that's not all that IS builds. Thanks for pointing that out.

What I meant to say, was that I'm not sure if we need that set of results in the background. Or maybe if we keep it, we make it `onSubmit` rather than `onStateChange`?

eunjae-lee:
@Meschreiber 
Okay, it's about the UX of the example then.

To rephrase the questions:

1. Results inside autocomplete
vs
2. Results as a separate result section outside autocomplete

2-1. If (2), update `onSubmit` vs `onStateChange`

I don't mind doing (1) but it will probably look similar to Codesandbox for 1 (birchbox).
If you think (1) is better, then I can make it show results when an item is clicked or enter is pressed on an item. (I guess that what you meant by onSubmit)

What do you think?

Meschreiber:
Hey @eunjae-lee , we definitely don't need 1) product results inside autocomplete display.

By using onSubmit I just meant something like this: https://codesandbox.io/s/wild-water-bjkpk, where the "background" results don't change on every state change, only when the user submits. Do you think that would be ok?

eunjae-lee:
> Hey @eunjae-lee , we definitely don't need 1) product results inside autocomplete display.
> 
> By using onSubmit I just meant something like this: https://codesandbox.io/s/wild-water-bjkpk, where the "background" results don't change on every state change, only when the user submits. Do you think that would be ok?

Yeah it's good to me. Just one little thing is, user has to click a suggestion and then press enter again to see the results. Should we add a little description under the search box? Or do you think it's good enough without it?

Meschreiber:
Ah, good point, that's definitely unintuitive. Probably best to either address it (add updating background results to `onSelect`?) or add a description like you said.

francoischalifour (approved):
I think the example looks good already! We can refine the behavior in another iteration, wdyt?

Meschreiber:
> Yup, I guess a simple description would work. What do you think?
> Can you write me a short sentence like.. "Please hit enter to display search results." or something like that? :)

Sounds good, will do as a suggestion!

Autocomplete/basic/index.html
```diff
@@ -0,0 +1,39 @@
...
+      rel="stylesheet"
+      href="https://cdn.jsdelivr.net/npm/instantsearch.css@7/themes/satellite-min.css"
+    />
+
+    <title>autocomplete-results-page</title>
+  </head>
+
+  <body>
+    <div class="container">
+      <h1>Basic Standalone Autocomplete</h1>
```
Meschreiber, line 27:
```suggestion
      <h1>Multi-source Autocomplete</h1>
```
I think "basic" and "standalone" might be confusing here. "Multi-source" is descriptive without being overkill (e.g., "Autocomplete with pre-defined items, recent searches, and suggested searches"). 

True "source" might be confusing, but a little more description below can explain that.

Autocomplete/basic/index.html
```diff
@@ -0,0 +1,39 @@
...
+      <p>Predefined Items + Recent Searches + Query Suggestions</p>
```
Meschreiber, line 28: WDYT of this for further explanation:

```suggestion
      <p>This autocomplete display items from multiple <a href="https://autocomplete.algolia.com/docs/sources/">sources</a>. It includes:</p>
      <ul>
        <li>
          pre-defined items, i.e., links to our Autocomplete Documentation and
          Github respository
        </li>
        <li>recent searches made by the user</li>
        <li>suggested searches powered by <a href="https://www.algolia.com/doc/guides/getting-insights-and-analytics/leveraging-analytics-data/query-suggestions/">Algolia's Query Suggestions feature<a></li>
      </ul>
      <p>
        You'll only see recent searches appear once you've submitted (pressed
        enter) on a query.
      </p>
      <p>To see results, hit enter once you've selected a suggested or recent search or typed a query.</p>
```

Meschreiber (commented):
WDYT of these language tweaks?

Autocomplete/basic/index.html
eunjae-lee: Makes sense!

Autocomplete/basic/index.html
eunjae-lee: Lovely :)

Autocomplete/basic/src/predefinedItemsPlugin.js
```diff
@@ -0,0 +1,44 @@
...
+          );
+        },
+        getItemUrl({ item }) {
+          return item.url;
+        },
+        templates: {
+          item({ item, root }) {
+            const content = document.createElement('a');
+            content.href = item.url;
+            content.className = 'aa-ItemContent aa-PredefinedItem';
```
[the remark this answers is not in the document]
eunjae-lee: Nice 👍 
https://github.com/algolia/doc-code-samples/pull/232/commits/8640b9f614708d48d0c4b5f8aba58b905976fbfc

Autocomplete/basic/index.html
```diff
@@ -0,0 +1,52 @@
...
+        <li>suggested searches powered by <a href="https://www.algolia.com/doc/guides/getting-insights-and-analytics/leveraging-analytics-data/query-suggestions/">Algolia's Query Suggestions feature<a></li>
+      </ul>
+      <p>
+        You'll only see recent searches appear once you've submitted (pressed
+        enter) on a query.
+      </p>
+      <p>To see results, hit enter once you've selected a suggested or recent search or typed a query.</p>
+      <div id="autocomplete"></div>
+      <p>
+        note: recent searches only get applied once you submit / press enter on
```
Meschreiber, line 44: This warning is now duplicated.

Meschreiber (changes requested):
The warning "note: recent searches only get applied once you submit / press enter on" is now duplicated. Other than that, LGTM! Will update guide(s) today or tomorrow! https://github.com/algolia/doc/pull/5033

Thanks again for your help and patience! 🙏

Autocomplete/basic/index.html
eunjae-lee: https://github.com/algolia/doc-code-samples/pull/232/commits/40971c28cc606ccdb8c5520390c0db3d831d7b45

Meschreiber (approved):
Thanks for your patience working through this!

PR #233, opened 2020-10-26
feat(autocomplete): update Query Suggestions to Autocomplete v1

francoischalifour:
This updates the [*Query Suggestions* guide](https://www.algolia.com/doc/guides/building-search-ui/resources/ui-and-ux-patterns/in-depth/query-suggestions/js/) to Autocomplete v1.

## Implementation

I use a child index that takes care of the Query Suggestions index. This index is "detached" in the sense that it doesn't refine the main index query on every key stroke, but only on select and on submit via the [`setUiState`](https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/#widget-param-setuistate) API. In addition to refining the search, it applies the categories from the selected item. On submit, we reset all categories.

## Preview

| Before | After |
|---|---|
| [Demo →](https://codesandbox.io/s/github/algolia/doc-code-samples/blob/master/InstantSearch.js/query-suggestions) | [Demo →](https://codesandbox.io/s/github/algolia/doc-code-samples/blob/feat/autocomplete-query-suggestions/InstantSearch.js/query-suggestions) |

## Notes

The Query Suggestions are not reverse-highlighted, but just highlighted with the [`highlight`](https://www.algolia.com/doc/api-reference/widgets/highlight/js/) helper. Instead of implementing this feature on the sandbox itself (and all the other ones), I believe we should support this in InstantSearch.js.

### DIFF
--- a/InstantSearch.js/query-suggestions/src/app.js
+++ b/InstantSearch.js/query-suggestions/src/app.js
@@ -84,13 +84,38 @@ function createAutocompleteSearchBox() {
                   return suggestion.query;
                 },
                 getSuggestions() {
-                  return currentIndex.hits;
+                  return currentIndex.hits.reduce((acc, current, i) => {
+                    const itemCategories = current.instant_search.facets.exact_matches.categories.map(
+                      x => x.value
+                    );
+
+                    const categories =
+                      i === 0 && itemCategories.length > 0
+                        ? [itemCategories[0]]
+                        : [];
+
+                    const itemWithoutCategories = {
+                      query: current.query,
+                      categories: [],
+                      ...current,
+                    };
+
+                    const items = categories.map(category => {
+                      return {
+                        query: current.query,
+                        categories: [category],
+                        ...current,
+                      };
+                    });
+
+                    acc.push(itemWithoutCategories, ...items);
+
+                    return acc;
+                  }, []);
                 },
                 onSelect({ suggestion }) {
                   setUiState(prevUiState => {
-                    const category = suggestion.instant_search.facets.exact_matches.categories.map(
-                      x => x.value
-                    )[0];
+                    const category = suggestion.categories[0];
 
                     return {
                       ...prevUiState,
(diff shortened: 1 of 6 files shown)


InstantSearch.js/query-suggestions/src/app.js
```diff
@@ -1,125 +1,191 @@
...
+import {
+  connectAutocomplete,
+  connectSearchBox,
+  connectRefinementList,
+} from 'instantsearch.js/es/connectors';
+import {
+  configure,
+  index,
+  hits,
+  refinementList,
```
Meschreiber, line 14:
```suggestion
```

Meschreiber: (Looks like only a custom refinementList is used.)

francoischalifour: Indeed—addressed in 55acdf4c5b045d3fde0ca35d908272de0909177d.

eunjae-lee (commented):
Hey, I just read the code and tested it. It looks fine to me.
And I was wondering what if we add a real refinement list, so that people can see the actual categories applied / removed as they use the autocomplete.
What do you think?

Meschreiber:
I like @eunjae-lee's suggestion above.

Also, as I was creating the gifs for the docs PR, I had a thought about displaying categories. I think a more practical UX might be one like Amazon presents, where the top suggestion offers a search without the category and then one with the category:

image

As is, each suggestion includes a category, and if you select it, the results aren’t actually that different (or not different at all) than if we did not include a category.

I think giving the users the ability to search “at large” or within a certain category makes more sense. (And that'll be nicely illustrated by having a not virtual refinementlist.) WDYT?

francoischalifour:
@eunjae-lee When I developed the sandbox I used an actual RefinementList: [see initial demo](https://codesandbox.io/s/autocomplete-instantsearchjs-and-categories-ye2c9)—but I believe this is for debugging purposes only with this display. A real-case example of that will be displaying a hierarchical menu instead of a refinement list. This allows the user to "skip" the initial "lvl0" refinement and to end up directly in the item's category. That sounds great and I'll update that!

@Meschreiber The experience you describe is the one I created for the Code Plays, and I'm fine using this one here as well: [see demo](https://codesandbox.io/s/autocomplete-code-plays-qs-and-rs-qfids?file=/src/index.tsx).

francoischalifour:
Here's the new experience in 190399b5efc4dd3047a670b33a4600483201c64c: 

image

A few notes:

- Ignore the messy markup and styles, Kevin and I are currently working on an Autocomplete theme that we'll use in the guides.
- The code might seem a little complex, let me know if you want to simplify it. Eventually, I see this QS experience as an Autocomplete Plugin, so very low code.

InstantSearch.js/query-suggestions/src/app.js
```diff
@@ -158,17 +171,21 @@ search.addWidgets([
     }),
     // We need to add a refinement list with the same attribute as the parent index
```
eunjae-lee, line 197:
```
     // We need to add a hierarchical menu with the same attribute as the parent index 
```

eunjae-lee (approved):
Looks good to me

InstantSearch.js/query-suggestions/src/app.js
francoischalifour: v.

francoischalifour:
I refactored the code to make it simpler. I believe we're good to move forward.

eunjae-lee (approved):
Yup, they're better indeed 👍

Shipow:
I've already share my concerns to @francoischalifour during the development of the feature but I'll quickly summarize my main concern about the integration with instantsearch.
Theautocompelte behavior and everything is very clean and fit perfectly in a tons of scenario but I think the refinement which is made with this input as a strong confidence of the user and then its state should not get that much importance in the UI afteward. In other words when you select a category on a marketplace/store your next move will unlikely be to move out to another category but refine your results. The way Amazon integrated a switcher/select to the mains categories and remove it from the search results page seems more interesting to free some important space/load on the rest of the flow.

francoischalifour:
@Shipow I believe what you're suggesting is to use a virtual hierarchical menu instead of the provided `hierarchicalMenu` widget. The hierarchical menu shouldn't show up but its refinement should still apply.

Do you think that a tradeoff would be to have an actual `hierarchicalMenu` in the guide, and to add a recommendation section at the end? This section would state that this widget doesn't need to be rendered if you display the main current search category at an obvious spot in your header.

Shipow:
yes, right that would work!

PR #235, opened 2020-12-10
fix(vue): update the wrong index name in the example

eunjae-lee:
# Summary

This PR updates the wrong index name (`demo_ecommerce` -> `instant_search`).
The vue-router example was broken.

* before: https://codesandbox.io/s/github/algolia/doc-code-samples/tree/master/Vue%20InstantSearch/routing-vue-router
* after: https://codesandbox.io/s/github/algolia/doc-code-samples/tree/fix/vue-router/Vue%20InstantSearch/routing-vue-router?file=/src/main.js

### DIFF
--- a/Vue InstantSearch/routing-vue-router/src/views/Home.vue
+++ b/Vue InstantSearch/routing-vue-router/src/views/Home.vue
@@ -87,17 +87,17 @@ export default {
         stateMapping: {
           stateToRoute(uiState) {
             return {
-              query: uiState.instant_search.query,
+              query: uiState.demo_ecommerce.query,
               brands:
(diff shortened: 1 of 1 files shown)


shortcuts (approved):
Not sure why the CI doesn't work but the changes looks good to me!

francoischalifour (commented):
We use the `instant_search` index everywhere, why are we using `demo_ecommerce` here?

eunjae-lee:
> We use the `instant_search` index everywhere, why are we using `demo_ecommerce` here?

I don't know. If that's the case, I'll update the example to use `instant_search` index.

francoischalifour (approved):

PR #239, opened 2021-03-08
feat(js): add an example for Segment integration

eunjae-lee:
## Summary

This PR adds an example for Segment integration in InstantSearch.js.

~The dependency is a temporary one at the moment. So we can wait until we release the new version of InstantSearch.js tomorrow (9 Mar) and reflect it here.~ → done


InstantSearch.js/segment-connector/index.html
```diff
@@ -0,0 +1,62 @@
...
+      analytics.load("YOUR_WRITE_KEY");
+      analytics.page();
+      }}();
+    </script>
+    <script src="https://cdn.jsdelivr.net/algoliasearch/3.32.0/algoliasearchLite.min.js"></script>
```
Haroenv, line 59: js client and InstantSearch imported via a different method?

InstantSearch.js/segment-connector/src/app.js
```diff
@@ -0,0 +1,128 @@
...
+  insightsClient,
+  onEvent(event) {
+    const { widgetType, eventType, payload, hits, attribute } = event;
+
+    console.log(event);
+    if (widgetType === 'ais.hits' && eventType === 'view') {
+      analytics.track('Product List Viewed', {
+        index: payload.index,
+        products: hits.map((hit) => ({
+          objectID: hit.objectID,
```
Haroenv, line 77: if we only needed the objectIDs, could it not have been read from the payload instead?

eunjae-lee: yes, but according to the [Segment spec](https://segment.com/docs/connections/spec/ecommerce/v2/#product-list-viewed), they will for sure put other properties, but I didn't to go too specific, so only put the necessary properties like that as a skeleton.

InstantSearch.js/segment-connector/index.html
eunjae-lee: https://github.com/algolia/doc-code-samples/pull/239/commits/ccc8ac085b99cfee9e142924ca38783bb98db242

and lowered the parcel version because it had an error https://stackoverflow.com/questions/66459081/parcel-semver-bug

InstantSearch.js/segment-connector/src/app.js
```diff
@@ -0,0 +1,127 @@
...
+
+// Set userToken to Segment
+analytics.identify('my-user-token');
+
+const insightsMiddleware = createInsightsMiddleware({
+  insightsClient: null,
+  onEvent(event) {
+    const { widgetType, eventType, payload, hits, attribute } = event;
+
+    console.log(event);
```
Haroenv, line 75: remove the log?

Haroenv (approved):

InstantSearch.js/segment-connector/src/app.js
eunjae-lee: I intentionally left it so that they see what's going on as soon as they run the example, but you think too much?

Haroenv: I think people will probably leave this log in their production sites, but that's probably not a big problem

PR #241, opened 2021-04-22
fix(conditional-request): spread algoliaClient

eunjae-lee:
## Summary

This PR spreads `algoliaClient` in `searchClient` in conditional requests examples.

We're mocking `searchClient` but only expose `search` function, so it can lead to unexpected error due to lack of other properties or methods.

In the recent support ticket, insights middleware failed to get appId and apiKey from the searchClient. So we'd better spread the algoliaClient to the searchClient.

rel: https://github.com/algolia/doc/pull/5697


Haroenv (approved):
this doesn't work for algoliasearch v3 (due to binding issues), but I think those cases are getting slim

francoischalifour (approved):

PR #244, opened 2021-07-07
first comit

Sabyh11:


Haroenv:
Thanks for your PR, but it seems like this code is for yourself, and thus doesn't need a pull request to change things on the official demos.

Hope you enjoy using Algolia!

PR #253, opened 2021-09-15
chore(CI): enable caching of node_modules for faster builds

tkrugg:


Haroenv (approved):
should be quicker

tkrugg:
I guess it's a little better -33% 

image

Issue #255, opened 2021-09-17
Upgrade Angular InstantSearch examples to Angular 12


tkrugg:
- [x] algolia-insights
- [x] autocomplete-results-page
- [x] conditional-request
- [x] debounced-search-box
- [x] extending-widgets
- [x] geo-search
- [x] getting-started
- [x] infinite-scroll
- [x] loading-indicator
- [x] ~manipulating-lists~
- [x] multi-index-autocomplete
- [x] multi-index-hits
- [x] ~places~ (deprecated)
- [x] query-suggestions
- [x] refresh
- [x] ~routing-angular-router~
- [x] routing-basic
- [x] routing-seo-friendly
- [x] routing-state-mapping
- [x] secured-api-keys

PR #256, opened 2021-09-17
chore(Angular InstantSearch): upgrade infinite-scroll example to Angular 12

tkrugg:

ref #255


Angular InstantSearch/infinite-scroll/src/app/app.component.html
```diff
@@ -1,37 +1,38 @@
...
-                                </li>
-                                <div (onVisible)="refine()"></div>
-                            </ol>
-                        </ng-template>
-                    </ais-infinite-hits>
-                </div>
+        <ais-infinite-hits>
+          <ng-template let-hits="hits" let-results="results" let-refine="showMore" >
+            <div *ngIf="hits.length === 0">
+              No results found matching <strong></strong>.
```
francoischalifour, line 24: The query is missing here. (It was already before this PR but it's worth either removing the message, or displaying the actual query.)

francoischalifour (approved):

Angular InstantSearch/infinite-scroll/src/app/app.component.html
tkrugg: Oh you're right, thanks for spotting that!

tkrugg:
```suggestion
              No results found.
```

PR #258, opened 2021-09-17
chore(Angular InstantSearch): upgrade multi-index-autocomplete example to Angular

tkrugg:

ref #255


francoischalifour (approved):
We should probably mention "Angular 12" in the commit name?

francoischalifour:
CodeSandbox doesn't seem to be able to load the example because there are too many dependencies. That's problematic since we rely on CodeSandbox for all our examples.

[See CodeSandbox deployment →](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/angular-12-multi-index-autocomplete)

tkrugg:
@francoischalifour codesandbox has always been a bit laky with Angular, but it works, usually if you open a private browser session. https://codesandbox.io/s/github/algolia/doc-code-samples/tree/angular-12-multi-index-autocomplete/Angular+InstantSearch/multi-index-autocomplete/?file=/tsconfig.json:340-366

PR #262, opened 2021-09-20
chore(Angular IS): upgrade refresh example to Angular 12

tkrugg:

ref #255


Angular InstantSearch/refresh/src/app/app.module.ts
```diff
@@ -1,14 +1,20 @@
...
-import { NgModule } from "@angular/core";
-import { NgAisModule } from "angular-instantsearch";
+import { BrowserModule } from '@angular/platform-browser';
+import { NgModule } from '@angular/core';
+import { NgAisModule } from 'angular-instantsearch';
 
-import { AppComponent } from "./app.component";
-import { Refresh } from "./refresh.component";
+import { AppComponent } from './app.component';
+import {RefreshComponent} from "./refresh.component";
```
francoischalifour, line 6: This line wasn't prettified?

Angular InstantSearch/refresh/src/app/refresh.component.ts
```diff
@@ -1,39 +1,75 @@
...
+          refresh() {},
+          widgetParams,
+          instantSearchInstance,
+        },
+        true
+      );
+    },
+    render({ instantSearchInstance }) {
+      const refresh = instantSearchInstance.refresh.bind(instantSearchInstance);
+      renderFn({ refresh, widgetParams, instantSearchInstance }, false);
```
francoischalifour, line 37: In the `init` method we use multiple lines for `renderFn` so we can do the same here for consistency.

francoischalifour (approved):

Angular InstantSearch/refresh/src/app/refresh.component.ts
tkrugg: fixed here: 
edf77ef

PR #263, opened 2021-09-21
chore(Angular IS): remove places example incompatible with Angular 12

tkrugg:
This removes the guide for places.
Places is depreacted: [EOL May 2022](https://www.algolia.com/blog/product/sunsetting-our-places-feature/)


Haroenv (approved):

Haroenv:
has the doc also been removed?

Haroenv:
why is the places still trying to build on CI @tkrugg ?

Haroenv (changes requested):
ci fails because something still is set to run it for angular

PR #269, opened 2021-10-11
feat(InstantSearch.js) Add e-commerce demo

clemfromspace:
Moved from https://github.com/algolia/instantsearch.js/tree/master/examples/e-commerce.


Haroenv (approved):
works!

PR #274, opened 2021-10-14
chore(Angular InstantSearch): upgrade secured-api-keys example to Angular 12

tkrugg:

ref #255


Angular InstantSearch/secured-api-keys/src/app/app.component.ts
```diff
@@ -1,19 +1,19 @@
-import { Component } from "@angular/core";
-import * as algoliasearch from "algoliasearch/lite";
+import { Component } from '@angular/core';
+import algoliasearch from 'algoliasearch/lite';
 
 const searchClient = algoliasearch(
-  "B1G2GM9NG0",
+  'B1G2GM9NG0',
   (window as any).SERVER_DATA.ALGOLIA_API_KEY
```
Haroenv, line 6: is the any still needed if it's in SERVER_DATA, angular doesn't expose that to the window interface? it's ok as-is though

Haroenv (approved):
🟢

Angular InstantSearch/secured-api-keys/src/app/app.component.ts
tkrugg: I'm not sure what you mean, we expose it manually [via the HTML we render server side](
https://github.com/algolia/doc-code-samples/pull/274/files#diff-ead47988dc5f8cd52133b62e150dda02e0c636d4fd6be5d707ee3f18c1f229c2R9-R11).

Or do you mean reading from window via a service with dependency injection?

francoischalifour: `SERVER_DATA` is a key that we created in the example ([see the guide](https://www.algolia.com/doc/guides/building-search-ui/going-further/api-keys-security/angular/#inline-the-api-key)). So Angular is unaware of this property on `window`.

PR #317, opened 2021-11-02
feat(react-instantsearch): update routing basic example

sarahdayan:
This updates the routing example in React InstantSearch:

- Uses hooks and functional components instead of classes
- Uses the latest version of the client

This change reflects the [documentation snippets](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/react/#basic-urls) (minor changes to do there, will handle) and solve a [query duplication issue](https://github.com/algolia/react-instantsearch/issues/3187).

fixes https://github.com/algolia/react-instantsearch/issues/3187

### DIFF
--- a/React InstantSearch/routing-basic/src/App.js
+++ b/React InstantSearch/routing-basic/src/App.js
@@ -33,7 +33,7 @@ export function App({ location, history }) {
     clearTimeout(debouncedSetStateRef.current);
 
     debouncedSetStateRef.current = setTimeout(() => {
-      history.push(searchStateToUrl(props, updatedSearchState));
+      history.push(searchStateToUrl(location, updatedSearchState));
     }, DEBOUNCE_TIME);
 
     setSearchState(updatedSearchState);
--- a/React InstantSearch/routing-basic/src/App.js
+++ b/React InstantSearch/routing-basic/src/App.js
@@ -20,13 +20,12 @@ const searchClient = algoliasearch(
 
 const createURL = state => `?${qs.stringify(state)}`;
 
-const searchStateToUrl = (props, searchState) =>
-  searchState ? `${props.location.pathname}${createURL(searchState)}` : '';
+const searchStateToUrl = (location, searchState) =>
+  searchState ? `${location.pathname}${createURL(searchState)}` : '';
 
 const urlToSearchState = location => qs.parse(location.search.slice(1));
 
-export function App(props) {
-  const { location, history } = props;
+export function App({ location, history }) {
   const [searchState, setSearchState] = useState(urlToSearchState(location));
   const debouncedSetStateRef = useRef(null);
 
--- a/React InstantSearch/routing-basic/src/App.js
+++ b/React InstantSearch/routing-basic/src/App.js
@@ -91,5 +91,3 @@ App.propTypes = {
   }),
   location: PropTypes.object.isRequired,
 };
-
-export default App;
--- a/React InstantSearch/routing-basic/src/App.js
+++ b/React InstantSearch/routing-basic/src/App.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useRef, useState } from 'react';
 import {
   InstantSearch,
   Hits,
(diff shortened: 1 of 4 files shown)


React InstantSearch/routing-basic/src/App.js
```diff
@@ -25,70 +25,58 @@ const searchStateToUrl = (props, searchState) =>
...
+
+  function onSearchStateChange(updatedSearchState) {
+    clearTimeout(debouncedSetState);
+
+    setDebouncedSetState(
+      setTimeout(() => {
+        history.push(
+          updatedSearchState,
+          null,
+          searchStateToUrl(updatedSearchState)
```
dhayab, line 40: `searchStateToUrl` expects two arguments (`props`, `searchState`), so in this case it will return an empty string.

React InstantSearch/routing-basic/src/App.js
```diff
@@ -25,70 +25,58 @@ const searchStateToUrl = (props, searchState) =>
...
+  const [searchState, setSearchState] = useState(urlToSearchState(location));
+  const [debouncedSetState, setDebouncedSetState] = useState(null);
+
+  function onSearchStateChange(updatedSearchState) {
+    clearTimeout(debouncedSetState);
+
+    setDebouncedSetState(
+      setTimeout(() => {
+        history.push(
+          updatedSearchState,
```
dhayab, line 38: I'm not very familiar with react-router, but looking at their API ref, `history.push` needs the path to be the first argument here.

sarahdayan: You're right, I didn't pay much attention because I thought the signatures hadn't changed from the past code (since I only updated the Algolia client).

Addressed in 24293ee.

React InstantSearch/routing-basic/src/App.js
sarahdayan: You're right, I didn't pay much attention because I thought the signatures hadn't changed from the past code (since I only updated the Algolia client).

Addressed in 24293ee.

dhayab (approved):

Haroenv (approved):

React InstantSearch/routing-basic/src/App.js
```diff
@@ -25,70 +25,55 @@ const searchStateToUrl = (props, searchState) =>
...
 
-class App extends Component {
-  state = {
-    searchState: urlToSearchState(this.props.location),
-    lastLocation: this.props.location,
-  };
+export function App(props) {
+  const { location, history } = props;
+  const [searchState, setSearchState] = useState(urlToSearchState(location));
+  const [debouncedSetState, setDebouncedSetState] = useState(null);
```
francoischalifour, line 31: This is not a state variable because we don't want to re-render when it changes. We can use a ref here.

francoischalifour (changes requested):
While we're at it, I think we can re-order the imports in `index.js`.

React InstantSearch/routing-basic/src/App.js
sarahdayan: Addressed in 5e9f823.

React InstantSearch/routing-basic/src/App.js
```diff
@@ -25,70 +25,53 @@ const searchStateToUrl = (props, searchState) =>
...
 
-class App extends Component {
-  state = {
-    searchState: urlToSearchState(this.props.location),
-    lastLocation: this.props.location,
-  };
+export function App(props) {
+  const { location, history } = props;
+  const [searchState, setSearchState] = useState(urlToSearchState(location));
+  const debouncedSetState = useRef(null);
```
francoischalifour, line 31:
```suggestion
  const debouncedSetStateRef = useRef(null);
```

React InstantSearch/routing-basic/src/App.js
```diff
@@ -25,70 +25,53 @@ const searchStateToUrl = (props, searchState) =>
...
-  onSearchStateChange = searchState => {
-    clearTimeout(this.debouncedSetState);
-
-    this.debouncedSetState = setTimeout(() => {
-      this.props.history.push(
-        searchStateToUrl(this.props, searchState),
-        searchState
-      );
+    debouncedSetState.current = setTimeout(() => {
+      history.push(searchStateToUrl(props, updatedSearchState));
```
francoischalifour, line 37: We just use `location` in `searchStateToUrl()`, so to make the example easier for users to modify, I don't think we should forward the whole `props` here (they would end up with other props if their `<App>` accept more props). It comes closer to the signature of `urlToSearchState()` too.

Suggestion:

```suggestion
      history.push(searchStateToUrl(location, updatedSearchState));
```

And then, we can destructure the props in `<App>`, as we're doing in `<Hit>`.

React InstantSearch/routing-basic/src/App.js
sarahdayan: Addressed in 6d94f94.

React InstantSearch/routing-basic/src/App.js
sarahdayan: Yep it makes sense, addressed in 96b2075.

francoischalifour (approved):

PR #320, opened 2021-11-08
fix(Angular InstantSearch): upgrade extending-widgets example

tkrugg:
Updates extending-widget example.
Related guide has already been updated: https://www.algolia.com/doc/guides/building-search-ui/widgets/customize-an-existing-widget/angular/#customize-the-complete-ui-of-the-widgets

https://github.com/algolia/doc-code-samples/issues/255


Angular InstantSearch/extending-widgets/src/app/app.component.html
```diff
@@ -1,30 +1,60 @@
...
+        <div class="searchbox">
+          <ais-search-box placeholder=""></ais-search-box>
+        </div>
+        <div class="filters">
+          <ais-menu-select></ais-menu-select>
+        </div>
+        <ais-hits>
+          <ng-template let-hits="hits" let-results="results">
+            <div *ngIf="hits.length === 0">
+              No results found matching <strong></strong>.
```
Haroenv, line 26:
```suggestion
              No results found.
```

we don't have access to the query here

Angular InstantSearch/extending-widgets/src/app/app.component.html
```diff
@@ -1,30 +1,60 @@
...
-          </ol>
-        </ng-template>
-      </ais-hits>
+    <div class="search-panel">
+      <div class="search-panel__results">
+        <div class="searchbox">
+          <ais-search-box placeholder=""></ais-search-box>
+        </div>
+        <div class="filters">
+          <ais-menu-select></ais-menu-select>
```
Haroenv, line 21: I know it's the same as the guide, but we probably should guide people to make their own prefix, not `ais`, right?

Haroenv (commented):
widget itself is good, two smaller questions :)

Angular InstantSearch/extending-widgets/src/app/app.component.html
tkrugg: That's an excellent point. I'll use `my-` prefix, WDYT ?

Haroenv: my- sounds good

Haroenv (approved):
obviously the guide also needs updating now :)

tkrugg:
@Haroenv 

> obviously the guide also needs updating now :)

Done here https://github.com/algolia/doc/pull/6430

PR #321, opened 2021-11-09
fix(Angular InstantSearch): upgrade geo-search example

tkrugg:
- used create-instantsearch-app to generate an Angular 12 boilerplate
- added [`@angular/google-maps`](https://github.com/angular/components/tree/master/src/google-maps) replacing [`@agm/core`](https://github.com/algolia/doc-code-samples/blob/master/Angular%20InstantSearch/geo-search/package.json#L14) which turned out to be incompatible with Angular 12

ref: https://github.com/algolia/doc-code-samples/issues/255


Angular InstantSearch/geo-search/package.json
```diff
@@ -1,54 +1,44 @@
...
-    "@angular/platform-browser-dynamic": "6.1.10",
-    "@angular/router": "6.1.10",
-    "algoliasearch": "3.33.0",
-    "angular-instantsearch": "3.0.0-beta.2",
-    "core-js": "2.6.7",
-    "instantsearch.js": "3.7.0",
-    "rxjs": "6.3.3",
-    "zone.js": "0.9.1"
+    "@agm/core": "^3.0.0-beta.0",
+    "@angular/animations": "~12.2.0",
```
Haroenv, line 14: do we want to keep these as ~?

Angular InstantSearch/geo-search/package.json
```diff
@@ -1,54 +1,44 @@
...
-    "@angular/platform-browser": "6.1.10",
-    "@angular/platform-browser-dynamic": "6.1.10",
-    "@angular/router": "6.1.10",
-    "algoliasearch": "3.33.0",
-    "angular-instantsearch": "3.0.0-beta.2",
-    "core-js": "2.6.7",
-    "instantsearch.js": "3.7.0",
-    "rxjs": "6.3.3",
-    "zone.js": "0.9.1"
+    "@agm/core": "^3.0.0-beta.0",
```
Haroenv, line 13:
```suggestion
```

Haroenv (commented):
👍

Angular InstantSearch/geo-search/package.json
tkrugg: I'm not strongly opinionated on this, but the advantage I see is that it generated less updates via dependabot. 
Or simple install, the .lock prevents it to break so no problem there. The risk is really when we add/remove things from the package.json and invalidate the .lock.
But since theses `~` come from Angular CLI itself, I think we can trust them and benefit from less dependabot PRs.

Angular InstantSearch/geo-search/src/styles.css
```diff
@@ -1,11 +1,3 @@
 /* You can add global styles to this file, and also import other style files */
-body,
-h1 {
-  margin: 0;
-  padding: 0;
-}
-
-body {
-  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica,
```
Haroenv, line 9: in the just-merged PR by @dhayab there's default css + the css installed via cdn. Which is the correct way?

tkrugg: fixed here: 7dd2c94b92aa9acb3d1ab3142b4b6d501676996c

Haroenv (approved):

PR #337, opened 2021-11-25
fix(React IS): fix back button behaviour with basic and seo friendly routing

tkrugg:
This change fixes the behaviour of the back button in basic and SEO friendly routing example
Prior to this change, the examples were missing the bit that updates the externally managed `searchState` when `location` changes:

```jsx
  const [searchState, setSearchState] = useState(urlToSearchState(location));

  useEffect(() => {
    setSearchState(urlToSearchState(location));
  }, [location]);
```

fixes [FX-246](https://algolia.atlassian.net/browse/FX-246)
Related doc PR: https://github.com/algolia/doc/pull/6465

### DIFF
--- a/React InstantSearch/routing-seo-friendly/src/App.js
+++ b/React InstantSearch/routing-seo-friendly/src/App.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import {
   InstantSearch,
   Hits,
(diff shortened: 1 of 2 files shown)


React InstantSearch/routing-seo-friendly/src/App.js
```diff
@@ -133,22 +133,27 @@ Hit.propTypes = {
   hit: PropTypes.object.isRequired,
 };
 
+let timeout;
```
francoischalifour, line 136: We should use a ref here instead of a global variable: https://github.com/algolia/doc-code-samples/blob/8a3ffbec4b1661c5a997e612e47a075e1c63a3be/React%20InstantSearch/routing-basic/src/App.js#L35-L37

React InstantSearch/routing-seo-friendly/src/App.js
```diff
@@ -133,22 +133,27 @@ Hit.propTypes = {
...
+
 const App = ({ location, history }) => {
   const [searchState, setSearchState] = useState(urlToSearchState(location));
-  const [debouncedSetState, setDebouncedSetState] = useState(null);
 
   const onSearchStateChange = updatedSearchState => {
-    clearTimeout(debouncedSetState);
+    if (timeout) {
+      clearTimeout(timeout);
+    }
```
sarahdayan, line 144: You don't need to make this conditional.

francoischalifour (approved):
We could probably even remove prop types in a next iteration.

Haroenv (approved):

sarahdayan (approved):

PR #341, opened 2022-01-06
feat(React InstantSearch): introduce injected content sample

sarahdayan:
## Summary

This introduces the injected content sample. The goal is to provide simple, generic examples to both illustrate the new [documentation guide](https://github.com/algolia/doc/pull/6541) and provide users with concrete but straightforward starters.

The demo contains several page to showcase different use cases, while reusing the same connector. It facilitates navigation while keeping the demo self-contained.

**Note:** the `connectInjectedHits.js` code was already reviewed twice (once in the RFC, once in the documentation PR). Feel free to ignore, unless you notice outstanding issues.

### Hits from another index

This page showcases injection of content that lives in another index. The content is injected at a fixed position.

Capture d’écran 2022-01-06 à 20 57 01

### Hits from a Rule

This page showcases injection of content that lives in a Rule, on the same index. The content is injected at a fixed position, but defined in the Rule.

Capture d’écran 2022-01-06 à 20 58 12

### Third-party hits

This page showcases injection of third-party content (not in Algolia). The content is injected at fixed positions.

The code uses a fake API with simulated latency so users can visualize how they'd fetch remote data.

Capture d’écran 2022-01-06 à 20 59 01

### Interspersed hits

This page showcases injection of non-adjacent hits, at a regular rhythm.

Capture d’écran 2022-01-06 à 21 03 14


React InstantSearch/injected-content/src/InjectedInfiniteHits.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { createClassNames, connectInfiniteHits } from 'react-instantsearch-dom';
+
+import { connectInjectedHits } from './connectInjectedHits';
+
+const cx = createClassNames('Hits');
+
+export const InjectedInfiniteHits = connectInfiniteHits(
```
Haroenv, line 8: or is it InfiniteInjectedHits?

React InstantSearch/injected-content/src/connectInjectedHits.js
```diff
@@ -0,0 +1,81 @@
...
+                ...acc,
+              ];
+            },
+            [hitFromMainIndex]
+          );
+      })
+      .flat();
+
+    return {
+      injectedHits,
```
Haroenv, line 78: in theory you could call this `hits` too, as it would override the prop passed by connectHits/connectInfiniteHits, meaning it doesn't need a different inner component anymore than hits/infiniteHits

Haroenv: (to be double-checked, and I'm also not sure if it's useful)

Haroenv (commented):
looks great!

React InstantSearch/injected-content/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "version": "1.0.0",
+  "private": true,
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "lint": "eslint .",
+    "lint:fix": "npm run lint -- --fix"
+  },
+  "dependencies": {
+    "algoliasearch": "3.35.1",
```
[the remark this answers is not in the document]
sarahdayan: Addressed in 8a1df00.

React InstantSearch/injected-content/src/InjectedInfiniteHits.js
sarahdayan: Could be both, I don't have a preference. We've called it `InjectedInfiniteHits` in the guide, so if we want to change, let's do it in both places after we merge.

React InstantSearch/injected-content/src/connectInjectedHits.js
sarahdayan: Traditional `hits` have a different shape so we might want to avoid it (just like we used `resultsByIndex` instead of `results`).

Haroenv (approved):

PR #345, opened 2022-01-11
feat(React InstantSearch): add Recommend implementation in injected content sample

dhayab:
This adds a **Recommend** page to the injected-content code sample, that will be used to illustrate a new in-depth section in the [**Injecting content between hits**](https://github.com/algolia/doc/pull/6541) guide.

[**Preview →**](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/feat/injected-content-recommend/React%20InstantSearch/injected-content?file=/src/pages/Recommend.js)

injecting-content-recommend

### DIFF
--- a/React InstantSearch/injected-content/src/App.css
+++ b/React InstantSearch/injected-content/src/App.css
@@ -132,26 +132,33 @@ em {
   display: block;
 }
 
-.ais-Hits-injected .Recommend {
+.Recommend .ais-Hits-item:first-of-type {
+  color: #00aeff;
+  border-color: currentColor;
+}
+
+.Recommend .ais-Hits-injected .FrequentlyBoughtTogether {
   color: #00aeff;
   background-color: #e5f1f7;
   border-color: currentColor;
   box-shadow: none;
 }
 
-.ais-Hits-injected .Recommend h2 {
+.Recommend .ais-Hits-injected .FrequentlyBoughtTogether h2 {
   margin-top: 0;
 }
 
-.ais-Hits-injected .Recommend ul {
+.Recommend .ais-Hits-injected .FrequentlyBoughtTogether ul {
   margin: 0;
   padding: 0;
   list-style: none;
-  display: flex;
+  flex-grow: 1;
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
   gap: 1rem;
 }
 
-.ais-Hits-injected .Recommend article {
+.Recommend .ais-Hits-injected .FrequentlyBoughtTogether article {
   margin: 0;
   background-color: #fff;
 }
@@ -177,6 +184,18 @@ em {
   }
 }
 
+@media (min-width: 768px) {
+  .Recommend .ais-Hits-item:first-of-type {
+    border-right: 0;
+  }
+
+  .Recommend .ais-Hits-injected .FrequentlyBoughtTogether {
+    margin-left: 0;
+    width: calc(100% / 4 * 3);
+    border-left: 0;
+  }
+}
+
 @media (max-width: 767px) {
   .ais-Hits-item {
     width: calc(50% - 1rem);
@@ -186,6 +205,19 @@ em {
   .ais-Hits-injected article {
     min-height: 250px;
     height: 100%;
+  }
+
+  .Recommend .ais-Hits-item:first-of-type {
+    border-bottom: 0;
+  }
+
+  .Recommend .ais-Hits-item:first-of-type div {
+    height: 200px;
+  }
+
+  .Recommend .ais-Hits-injected .FrequentlyBoughtTogether {
+    margin-top: 0;
+    border-top: 0;
   }
 
   .sm\:Block-full {
--- a/React InstantSearch/injected-content/src/App.css
+++ b/React InstantSearch/injected-content/src/App.css
@@ -132,6 +132,30 @@ em {
   display: block;
 }
 
+.ais-Hits-injected .Recommend {
+  color: #00aeff;
+  background-color: #e5f1f7;
+  border-color: currentColor;
+  box-shadow: none;
+}
+
+.ais-Hits-injected .Recommend h2 {
+  margin-top: 0;
+}
+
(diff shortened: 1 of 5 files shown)


React InstantSearch/injected-content/src/pages/Recommend.js
```diff
@@ -0,0 +1,103 @@
...
+    </InstantSearch>
+  );
+}
+
+function RelatedProducts({ resultsByIndex }) {
+  const { recommendations } = useFrequentlyBoughtTogether({
+    recommendClient,
+    indexName: 'test_FLAGSHIP_ECOM_recommend',
+    objectIDs: resultsByIndex.test_FLAGSHIP_ECOM_recommend.hits
+      .slice(0, 2)
```
Haroenv, line 61: why not related to more than the first two items? this seems on purpose, but I don't really understand why

React InstantSearch/injected-content/src/pages/Recommend.js
```diff
@@ -0,0 +1,103 @@
...
+        height: '100%',
+        flexDirection: 'column',
+        justifyContent: 'space-around',
+      }}
+    >
+      <img
+        src={hit.image_urls[0]}
+        alt={hit.name}
+        style={{
+          width: '100%',
```
Haroenv, line 94: I think as it has width 100% it causes moments where the page shifts size when the images don't load simultaneously (then one takes up more space than the other) and between replacing the images:


https://user-images.githubusercontent.com/6270048/149111589-704b4db3-328a-4d54-b2a4-90365471e172.mov

I wonder if grid avoids this non-simultaneous size changing? Only if we always have the same amount of items of course, when recommend is loading, we need to make sure to render two placeholders too

Haroenv (commented):
I think this experience can be improved somewhat

React InstantSearch/injected-content/src/pages/Recommend.js
sarahdayan: I believe it would break the grid otherwise.

React InstantSearch/injected-content/src/App.css
```diff
@@ -132,6 +132,30 @@ em {
...
+
+.ais-Hits-injected .Recommend h2 {
+  margin-top: 0;
+}
+
+.ais-Hits-injected .Recommend ul {
+  margin: 0;
+  padding: 0;
+  list-style: none;
+  display: flex;
```
sarahdayan, line 150:
```suggestion
  flex-grow: 1;
  display: grid;
  grid-template-columns: 1fr 1fr;
```

React InstantSearch/injected-content/src/pages/Recommend.js
sarahdayan: We should be able to avoid this effect by doing the change suggested in `App.css` and adding placeholders as @Haroenv suggested. It can be as simple as setting a wrapping `<div>` with a defined size and a background color.

React InstantSearch/injected-content/src/pages/Recommend.js
Haroenv: Sorry, I wasn't asking why there are two results, that makes sense, I was asking why it's related to the first two hits only

sarahdayan: Ah! Yes that's a good question.

francoischalifour:
This example proves that Recommend works as injected content, but the example feature and UI is confusing.

1. Why compute the FBT on the two first results when they can be unrelated?
1. The FBT bundle is not visually linked to the reference items

I would suggest to use the first item as a reference, and to display 3 FBT to increase the selection choice.

image

Here's a [sandbox](https://codesandbox.io/s/busy-lumiere-x8zub?file=/src/pages/Recommend.js) with the code changes (the grid CSS needs to be reworked).

Wdyt?

dhayab:
@francoischalifour I loosely followed [the recommendation](https://algolia.slack.com/archives/C02PN3G4SAF/p1639762874011200) on the earlier POC. Your suggestion makes sense 👍, I'll make the changes.

React InstantSearch/injected-content/src/pages/Recommend.js
dhayab: Thanks for the suggestion, that should work better now.

React InstantSearch/injected-content/src/pages/Recommend.js
dhayab: Answered to @francoischalifour below. Now the recommendations are computed for the first item and are visually linked to it.

sarahdayan:
@dhayab This was in a private conversation, adding this here for visibility:

from @seafoox:

>Chatting with Marie, she suggested that if we do explore this direction, her recommendation would be to only pass the first few hits’ objectID to Recommend, and retrieve/display max 1-2 recommendations as part of the slot.

sarahdayan:
New UI looks cool! Careful to handle cases when there might be less than 3 recommendations sometimes though.

Capture d’écran 2022-01-14 à 11 34 46

React InstantSearch/injected-content/src/App.css
```diff
@@ -132,6 +132,38 @@ em {
...
+  margin-top: 0;
+}
+
+.Recommend .ais-Hits-injected .FrequentlyBoughtTogether ul {
+  margin: 0;
+  padding: 0;
+  list-style: none;
+  flex-grow: 1;
+  display: grid;
+  grid-auto-flow: column;
```
sarahdayan, line 158: This way, each colum takes exactly the same space even with longer text.

```suggestion
  grid-auto-flow: column;
  grid-auto-columns: 1fr;
```

sarahdayan (approved):
One tiny adjustment and we're good.

Congratulations on this one @dhayab 🔥

dhayab:
It appears to me it will be difficult to accomodate 3-column grid layout in mobile, so I only kept it on the desktop media query. On mobile it now displays FBT products one by one.

Haroenv (approved):
perfect!

PR #346, opened 2022-01-12
Update package.json

CatAlgolia:
add main parameter so codesandbox builds


Haroenv (changes requested):
you're right that this is the error message, but this isn't why sandbox doesn't work; as it simply doesn't seem to find the files at https://codesandbox.io/s/github/algolia/doc-code-samples/tree/master/React%20InstantSearch/routing-seo-friendly at all. Adding a main did not help IME

Haroenv:
Tracked in https://github.com/codesandbox/codesandbox-client/issues/6388

PR #370, opened 2022-03-14
Updated the Vue IS geosearch demo to use the latest version

seafoox:
Previously it was running on Vue InstantSearch v3.x, and the showing in the demo's title v2. Removed that reference and bumped to the latest VueIS version. Bonus: fixed the name of the package.

### DIFF
--- a/Vue InstantSearch/geo-search/package.json
+++ b/Vue InstantSearch/geo-search/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "infinite-scroll",
+  "name": "geosearch-with-googlemaps",
   "version": "0.1.0",
   "private": true,
   "scripts": {
@@ -12,7 +12,7 @@
     "instantsearch.css": "7.3.1",
     "vue": "2.6.10",
     "vue-googlemaps": "0.1.2",
-    "vue-instantsearch": "3.8.1"
+    "vue-instantsearch": "latest"
   },
   "devDependencies": {
     "@vue/cli-plugin-babel": "3.6.0",
--- a/Vue InstantSearch/geo-search/src/App.vue
+++ b/Vue InstantSearch/geo-search/src/App.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="root">
     <header class="header">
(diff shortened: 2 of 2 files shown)


Vue InstantSearch/geo-search/package.json
```diff
@@ -12,7 +12,7 @@
     "instantsearch.css": "7.3.1",
     "vue": "2.6.10",
     "vue-googlemaps": "0.1.2",
-    "vue-instantsearch": "3.8.1"
+    "vue-instantsearch": "latest"
```
Haroenv, line 15: this shouldn't use latest, but update manually to the latest version. This way it won't break when we release a major version

Haroenv: this change in the package.json needs to be reflected in yarn.lock

Haroenv (changes requested):
👍

Vue InstantSearch/geo-search/package.json
seafoox: Understood, good to know.
As breaking changes are expected to happen only on major version change, would it be relevant to use ^4.0.0 here, so  both upgrade patch and minor version updates will be pulled?

Haroenv: yes

seafoox: I've done the change, is that good for you now?

Vue InstantSearch/geo-search/package.json
```diff
@@ -12,7 +12,7 @@
     "instantsearch.css": "7.3.1",
     "vue": "2.6.10",
     "vue-googlemaps": "0.1.2",
-    "vue-instantsearch": "3.8.1"
+    "vue-instantsearch": "^3.8.1"
```
Haroenv, line 15:
```suggestion
    "vue-instantsearch": "^4.3.3"
```

3.8.1 isn't the latest :) If updated across the major version, you need to check whether the imports still work (they normally should be the same, unless possibly /es replaced by /vue2/es)

Haroenv (commented):
either change to 3.9.0 or 4.3.3

PR #375, opened 2022-04-01
fix(instantsearch): e-commerce

tkrugg:
Not sure it makes sense to have thses examples here and maintain them in two places.
In the meantime, this one broken because it wasn't build as part of the CI.


Haroenv:
it's here as @clemfromspace asked for it, for code sharing

tkrugg:
I'm not crazy about having to maintain theses in two places. I'll suggest to them to link it directly from the flavours repos and ultimately remove them from here. I think this could apply to all e-commerce & media demos for all flavours (8 examples)

sarahdayan (approved):
Looks fine 👍

Let's settle on demo/examples strategy in a dedicated meeting.

PR #388, opened 2022-06-08
feat(React InstantSearch Hooks): introduce multi-index hits sample

sarahdayan:
This introduces a multi-index sample for React InstantSearch Hooks.


React InstantSearch Hooks/multi-index-hits/src/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+import App from './App';
+
+import './index.css';
+
+const container = document.getElementById('root');
+const root = createRoot(container);
+
+root.render(
+  <React.StrictMode>
```
Haroenv, line 12: I thought we still had timing issues with strict mode, this works?

francoischalifour: It only breaks routing since [v6.27.0](https://github.com/algolia/react-instantsearch/releases/tag/v6.27.0). But I'm not sure we should enable it by default until we officially support it.

francoischalifour (approved):

PR #391, opened 2022-06-11
feat(React InstantSearch Hooks): introduce conditional debouncing sample

sarahdayan:
This introduces a conditional debouncing sample for React InstantSearch Hooks using the [Network Information API](https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API).


React InstantSearch Hooks/conditional-debouncing/src/App.js
```diff
@@ -0,0 +1,50 @@
...
+
+function App() {
+  return (
+    <div className="container">
+      <InstantSearch indexName="instant_search" searchClient={searchClient}>
+        <SearchBox
+          queryHook={(query, search) => {
+            const timeout = navigator.connection?.effectiveType?.includes('2g')
+              ? 400
+              : 0;
```
FabienMotte, line 27:
```suggestion
            const effectiveType = navigator.connection?.effectiveType;
            const rtt = navigator.connection?.rtt;
            const saveData = navigator.connection?.saveData;

           const timeout = effectiveType?.includes('2g') || rtt > 500 || saveData
             ? 400
             : 0;
```

Should we also base the condition on the [`navigator.connection.rtt`](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/rtt) and [`navigator.connection.saveData`](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/saveData) values?

sarahdayan: `effectiveType` already computes the `rtt` so we don't need to include it.

`saveData` can definitely be interesting if we know we're on a cellular network, or tethering to one. There's the `type` attribute but it's only available on Chrome OS so I'm not sure it's worth the hassle.

Haroenv: why do you think effectiveType is better to use than a number? I'd prefer rtt or downlink too

sarahdayan: As mentioned on the related guide, `effectiveType` is already a combination of `rtt` + `downlink`. [See MDN](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/effectiveType):
>This value is determined using a combination of recently observed round-trip time and downlink values.

Using `rtt` and `downlink` basically means recomputing our own `effectiveType` with our own arbitrary values. I think using `effectiveType` already does this job for us and is simpler for a code sample.

Haroenv (approved):
I personally think it makes more sense to calculate it only once, but this is fine too, and it assumes ordering in the effectiveType string, which isn't there necessarily (think 12G)

Haroenv:
good idea with a strict list :) I'm just not sure if getting the effectiveType property is as cheap as getting a key from any object, or if it does calculation, because in that case it might not be appropriate to check on every keypress

sarahdayan:
@Haroenv It's not as fast, no. We could also listen to changes on the `connection` object and update some local state for the `timeout`. I initially did this, but there was some funky behavior with `queryHook` and re-renders (likely because of unstable functions references). I can investiagte further if you want.

dhayab (approved):

React InstantSearch Hooks/conditional-debouncing/src/App.js
```diff
@@ -0,0 +1,63 @@
...
+import './App.css';
+
+const searchClient = algoliasearch(
+  'latency',
+  '6be0576ff61c053d5f9a3225e2a90f76'
+);
+
+const connection = navigator.connection;
+let timerId = undefined;
+let timeout = getTimeout();
```
Haroenv, line 20:
```suggestion
let timeout = 0;
updateTimeout()
```

React InstantSearch Hooks/conditional-debouncing/src/App.js
```diff
@@ -0,0 +1,63 @@
...
+  timerId = setTimeout(() => search(query), timeout);
+}
+
+function updateTimeout() {
+  timeout = getTimeout();
+}
+
+function getTimeout() {
+  return ['slow-2g', '2g'].includes(connection?.effectiveType) ? 400 : 0;
+}
```
Haroenv, line 53:
```suggestion
function updateTimeout() {
  timeout = ['slow-2g', '2g'].includes(connection?.effectiveType) ? 400 : 0;
}
```

Haroenv (approved):
up to you to choose, but it can be simplified a bit

dhayab (approved):

PR #392, opened 2022-06-15
fix(hooks): update Insights middleware with `useInstantSearch()`

francoischalifour:
This updates the Insights example to use `useInstantSearch()` introduced in v6.28.0.

We don't use `useEffect()` but rather `useLayoutEffect()` to trigger a single network request on initial load.


francoischalifour:
I updated the docs in https://github.com/algolia/doc/pull/7108.

PR #396, opened 2022-06-28
feat(hooks): document query suggestions

dhayab:
Code sample for https://github.com/algolia/doc/pull/7152.


React InstantSearch Hooks/query-suggestions/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "build": "parcel build index.html"
+  },
+  "dependencies": {
+    "@algolia/autocomplete-js": "^1.7.0",
+    "@algolia/autocomplete-plugin-query-suggestions": "^1.7.0",
+    "@algolia/autocomplete-theme-classic": "^1.7.0",
+    "algoliasearch": "4",
+    "instantsearch.js": "4.40.5",
+    "react": "17.0.2",
+    "react-autosuggest": "^10.1.0",
```
sarahdayan, line 16: I don't think we use it.

React InstantSearch Hooks/query-suggestions/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "react-dom": "17.0.2",
+    "react-instantsearch-hooks-web": "6.29.0"
+  },
+  "devDependencies": {
+    "@types/react": "17.0.45",
+    "@types/react-autosuggest": "^10.1.5",
```
sarahdayan, line 22: I don't think we use it.

React InstantSearch Hooks/query-suggestions/src/Autocomplete.tsx
```diff
@@ -0,0 +1,76 @@
...
+        return {
+          ...source,
+          onSelect({ item }) {
+            setIndexUiState((indexUiState) => ({
+              ...indexUiState,
+              query: item.query,
+              menu: {
+                categories: item.__autocomplete_qsCategory || '',
+              },
+            }));
```
sarahdayan, line 52: I would probably forward Autocomplete props to the component instead of doing it all inside the component, to keep concerns separated.

This way you don't use Hooks directly inside `<Autocomplete>` or do any specific logic. If the user wants to reuse the component differently somewhere else, they can without duplicating it.

React InstantSearch Hooks/query-suggestions/src/App.css
```diff
@@ -0,0 +1,23 @@
...
+  max-width: 1200px;
+  margin: 0 auto;
+  padding: 1rem;
+}
+
+.current-refinements {
+  display: flex;
+  align-items: center;
+  padding: 1rem 0;
+  min-height: 64px;
```
sarahdayan, line 22: Is this really necessary? Disabling it seems to work just fine.

sarahdayan (commented):
We also need to deal with small screens. You can check the CSS for recent samples such as [`conditional-debouncing`](https://github.com/algolia/doc-code-samples/blob/3c13f60d2479bde377415f7bd4a1158520500141/React InstantSearch Hooks/conditional-debouncing/src/App.css#L1) for an example.

Capture d’écran 2022-06-29 à 09 54 21

React InstantSearch Hooks/query-suggestions/src/App.css
dhayab: It prevents a subtle layout shift when refinements are added / cleared.

CleanShot 2022-06-30 at 15 32 27

React InstantSearch Hooks/query-suggestions/src/Autocomplete.tsx
dhayab: I fear it would make the sample a bit more complicated.

Since we can't use Hooks on the same component that renders `<InstantSearch>` (App.tsx here), it means we would need to create another layer just to decouple logic from the Autocomplete component.

Do you think we should go ahead with that nevertheless?

sarahdayan: Yup that's a trade-off, up to you if you think this adds too much complexity.

If we go this route, I think you also don't need to pass the search client as a prop. You can initialize it separately and import it, that should further reduce the noise in app code.

dhayab: I've integrated your proposal in 70983fe

React InstantSearch Hooks/query-suggestions/src/QuerySuggestions.tsx
```diff
@@ -0,0 +1,56 @@
...
+      placeholder="Search for products..."
+      plugins={[querySuggestionsPlugin]}
+      initialState={{ query }}
+      onReset={() =>
+        setIndexUiState((setIndexUiState) => ({
+          ...setIndexUiState,
+          query: '',
+          menu: { categories: '' },
+        }))
+      }
```
francoischalifour, line 53: We'd need to memo each of these props so that it doesn't rerun the effect in `<Autocomplete>`.

React InstantSearch Hooks/query-suggestions/src/QuerySuggestions.tsx
```diff
@@ -0,0 +1,56 @@
...
+import {
+  useInstantSearch,
+  useMenu,
+  useSearchBox,
+} from 'react-instantsearch-hooks-web';
+
+import { Autocomplete } from './Autocomplete';
+import { searchClient } from './searchClient';
+
+export function QuerySuggestions() {
```
francoischalifour, line 12: The name `QuerySuggestions` doesn't imply that it mounts a search box.

Alternative suggestions:

- AutocompleteSearchBox
- SearchBoxWithSuggestions

React InstantSearch Hooks/query-suggestions/src/Autocomplete.tsx
```diff
@@ -0,0 +1,31 @@
...
+    const instance = autocomplete({
+      ...props,
+      container: container.current,
+      renderer: { createElement, Fragment, render },
+    });
+
+    return () => instance.destroy();
+  }, [props]);
+
+  return <div ref={container} />;
```
dhayab, line 30:
```suggestion
  return <div ref={containerRef} />;
```

React InstantSearch Hooks/query-suggestions/src/QuerySuggestions.tsx
dhayab: I went with `SearchBoxWithSuggestions`, thanks for the suggestions!

dhayab:
Hi @sarahdayan @francoischalifour I applied your latest suggestions regarding props memoization, let me know if that works for you 🙇 .

francoischalifour (approved):

sarahdayan (approved):

PR #397, opened 2022-06-30
fix: remove spaces in directories

dhayab:
A possible cause of the recurrent instability issues we have with CodeSandBox could be related to spaces in our folders.

This PR renames all of the first-level directories to remove spaces. ~For consistency, "InstantSearch.js" is also renamed to "instantsearch-js" as it has been mentioned that the `.js` part could be problematic as well (I don't recall where it was discussed though).~ (i mixed this with a recent issue regarding different imports in react 17 / 18)

This PR should be merged in sync with its [counterpart PR in doc](https://github.com/algolia/doc/pull/7176).


.circleci/config.yml
```diff
@@ -37,148 +37,148 @@ jobs:
...
-           path: "React InstantSearch/algolia-insights"
+           path: "react-instantsearch/algolia-insights"
 
   build_vanilla_projects:
     executor: node14
     steps:
       - checkout
       - build_project:
-           path: "InstantSearch.js/autocomplete-results-page"
+           path: "instantsearch-js/autocomplete-results-page"
```
francoischalifour, line 110: We could use the package name instead? (`instantsearch.js`)

francoischalifour (approved):
Good to merge once docs are deployed.

francoischalifour:
All the previously linked sandboxes won't work anymore I believe though. Is this problematic?

dhayab:
@francoischalifour we've discussed this a bit [on Slack](https://algolia.slack.com/archives/C02E28JP9V2/p1656591529277619), and I think as long we're updating the references in our doc and other references we own it should be fine.

Haroenv (approved):
lgtm once documentation has a sync PR

PR #399, opened 2022-07-06
feat(hooks): document geo search

dhayab:
Code sample for a guide demonstrating how to use geo search capabilities with React InstantSearch Hooks.

Ticket: [FX-1494](https://algolia.atlassian.net/browse/FX-1494)
Doc PR : TBD

[**Sandbox →**](https://codesandbox.com/s/github/algolia/doc-code-samples/tree/feat/hooks-geo-search/react-instantsearch-hooks/geo-search)


react-instantsearch-hooks/geo-search/sandbox.config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "template": "node",
+  "container": {
+    "port": 1234
+  }
+}
```
dhayab, line 6: I'm shipping this config to rely on `parcel` in Codesandbox instead of their internal mechanism that currently has issues with type imports in TypeScript and automatic JSX runtime (no explicit `import React from 'react'`).

react-instantsearch-hooks/geo-search/sandbox.config.json
```diff
@@ -0,0 +1,6 @@
+{
+  "template": "node",
```
Haroenv, line 2: backend containers are much slower, could we instead use code that works on codesandbox too (without automatic runtime)?

react-instantsearch-hooks/geo-search/src/Airports.tsx
```diff
@@ -0,0 +1,86 @@
...
+    if (currentRefinement) {
+      clearMapRefinement();
+    }
+
+    if (items.length > 0 && !skipViewEffect.current) {
+      map.setView(items[0]._geoloc);
+    }
+
+    skipViewEffect.current = false;
+  }, [query]);
```
Haroenv, line 60: React Hook useEffect has missing dependencies: 'clearMapRefinement', 'currentRefinement', 'items', and 'map'. Either include them or remove the dependency array. (react-hooks/exhaustive-deps)eslint

should this be fixed or ignored?

react-instantsearch-hooks/geo-search/src/Airports.tsx
```diff
@@ -0,0 +1,86 @@
...
+    currentRefinement,
+    clearMapRefinement,
+  } = useGeoSearch();
+
+  const onViewChange = ({ target }) => {
+    skipViewEffect.current = true;
+
+    if (query.length > 0) {
+      refineQuery('');
+    }
```
Haroenv, line 37: if you only zoom, this also gets applied, however the searchbox doesn't clear, do you have any idea what happened?

Haroenv (commented):
great attention to detail!

react-instantsearch-hooks/geo-search/sandbox.config.json
dhayab: I thought about it and [I'm removing this sandbox config file](https://github.com/algolia/doc-code-samples/pull/399#discussion_r915904516) as there are fewer gains than I thought.

react-instantsearch-hooks/geo-search/src/Airports.tsx
```diff
@@ -0,0 +1,75 @@
...
+    if (currentRefinement) {
+      clearMapRefinement();
+    }
+
+    if (items.length > 0 && !skipViewEffect.current) {
+      map.setView(items[0]._geoloc);
+    }
+
+    skipViewEffect.current = false;
+  }, [query]);
```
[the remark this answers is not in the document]
dhayab: The original intent is:

- to center the map on the first item when the query changes
- to refine on the viewable part of the map when it is moved / zoomed

The first part requires me to reset the boundaries refinement if necessary (`clearMapRefinement()`) when the query changes to allow searching for airports that are not in the currently viewable portion of the map.

The last part requires me to reset the query otherwise it will refine on results that match the query on top of the geographic coordinates. And so I needed to mark with a ref (`skipViewEffect`) this particular action (resetting the query due to the map being moved) in order not to recenter the map on the first item of a results list.

react-instantsearch-hooks/geo-search/src/Airports.tsx
```diff
@@ -0,0 +1,75 @@
...
+    currentRefinement,
+    clearMapRefinement,
+  } = useGeoSearch();
+
+  const onViewChange = ({ target }) => {
+    skipViewEffect.current = true;
+
+    if (query.length > 0) {
+      refineQuery('');
+    }
```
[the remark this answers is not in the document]
dhayab: I explained the process [here](https://github.com/algolia/doc-code-samples/pull/399/files?diff=unified&w=0#r917957294), but in a gist, moving the map will trigger a refinement on its boundaries, and having the query will filter the results on this additional parameter.

This is not ideal when searching for airports by moving the map, when a user would expect all of them to be shown instead of those matching a preset query.

react-instantsearch-hooks/geo-search/src/Airports.tsx
dhayab: It is normally applied, is it still the case for you in the [latest preview](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/feat/hooks-geo-search/react-instantsearch-hooks/geo-search) ?

react-instantsearch-hooks/geo-search/src/Airports.tsx
dhayab: I removed the need for `useEffect` since now the behavior is done during render.

react-instantsearch-hooks/geo-search/src/Airports.tsx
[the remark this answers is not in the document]
francoischalifour: I think it's fine as long as we add comments to the code here, since it's complex and it's supposed to be understandable by anyone.

react-instantsearch-hooks/geo-search/src/Airports.tsx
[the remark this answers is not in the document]
francoischalifour: I see. Ideally the query should just be used to center the map, and then the query is not sent to Algolia anymore. That's my interpretation of what Google Maps does.

react-instantsearch-hooks/geo-search/src/Airports.tsx
Haroenv: maybe here the input can be blurred first

Haroenv (approved):

PR #400, opened 2022-07-07
fix(React Native): update `<SearchBox>` without Effects

francoischalifour:
## Description

This removes the `useEffect()`s from the `<SearchBox>` component in the React Native example.

## Related

- https://github.com/algolia/doc/pull/7189

### DIFF
--- a/react-instantsearch-hooks-native/getting-started/src/SearchBox.tsx
+++ b/react-instantsearch-hooks-native/getting-started/src/SearchBox.tsx
@@ -8,35 +8,30 @@ type SearchBoxProps = UseSearchBoxProps & {
 
 export function SearchBox({ onChange, ...props }: SearchBoxProps) {
   const { query, refine } = useSearchBox(props);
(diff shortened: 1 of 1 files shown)


kopax-polyconseil:
Hello @francoischalifour , could you please explain why you removed this `useEffect`  hook ? We currently use the hook version.

francoischalifour:
@kopax-polyconseil I explain in algolia/react-instantsearch#3551:

> You can learn more in [**You Might Not Need an Effect**](https://beta.reactjs.org/learn/you-might-not-need-an-effect), notably these two sections:
> 
> - [Adjusting some state when a prop changes](https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)
> - [Sharing logic between event handlers](https://beta.reactjs.org/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers)

PR #407, opened 2022-10-11
fix(hooks): better parity with the IS.js flavor code sample

FabienMotte:
This PR adds missing custom CSS styles, a `classNames.root` property, and renames the `Dropdown` component to `FacetDropdown` to avoid potential conflicts with user code.


FabienMotte:
The doc PR is still waiting for approval: https://github.com/algolia/doc/pull/7333

PR #408, opened 2022-10-12
fix(js): correct sample for multi-index-hits

Haroenv:
This has been changed a long time ago in https://www.algolia.com/doc/guides/building-search-ui/ui-and-ux-patterns/multi-index-search/js/, but for some reason never updated here

FX-1802

### DIFF
--- a/instantsearch.js/multi-index-hits/src/app.js
+++ b/instantsearch.js/multi-index-hits/src/app.js
@@ -5,34 +5,12 @@ const searchClient = algoliasearch(
   '6be0576ff61c053d5f9a3225e2a90f76'
 );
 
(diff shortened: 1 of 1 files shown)


dhayab (approved):
Woah, didn't know multi-index search was done like that before the `index()` widget.

I'm wondering if we could maybe switch to html tagged templates in this code sample, even if it doesn't exactly match what's in the guide. Wdyt?

Haroenv:
> I'm wondering if we could maybe switch to html tagged templates in this code sample, even if it doesn't exactly match what's in the guide. What do you think?

Let's do that in a separate PR, because it should be done in all repos

PR #412, opened 2022-10-20
feat(js): add content-injection demo

Haroenv:
see https://github.com/algolia/instantsearch.js/issues/4999

This supports
- infinite hits
- injection of any index
- injection of any rule

this doesn't yet support (easily)
- injection of third-party results
- complete typescript typing


Haroenv:
It didn't cover all the use cases we envisioned when writing the guide in React InstantSearch, but it still mostly works.

Feel free to share this anywhere, but it would require more effort on our case to define whether third party requests make sense to skip.

PR #419, opened 2022-11-08
fix(js): call `start` on correct object

sarahdayan:
A bug sneaked into #408 causing the example to no longer work.

This fixes it by calling the `start` function on the right object.

In the future, we could fix this by using TypeScript as much as possible, and testing examples.


Haroenv (approved):

Haroenv:
examples are already tested, but indeed only that they build, nothing else. Thanks for catching!

The reason most examples aren't in typescript is because we noticed quite a lot of people being "scared" as soon as they see the extension, but maybe that's a risk worth taking (ensuring we keep the type syntax to a bare minimum)

aymeric-giraudet (approved):

PR #421, opened 2022-11-08
fix(hooks): add and set up TypeScript

sarahdayan:
The [Query Suggestions](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/master/react-instantsearch-hooks/query-suggestions) example fails on CodeSandbox because of the `import type` statement and the fact that we don't import React in TSX files.

TypeScript is actually missing from dependencies and there is no `tsconfig.json` file. This PR adds them to fix the problems.

### DIFF
--- a/react-instantsearch-hooks/query-suggestions/tsconfig.json
+++ b/react-instantsearch-hooks/query-suggestions/tsconfig.json
@@ -1,0 +1,11 @@
+{
+  "compilerOptions": {
+    "target": "ESNext",
+    "module": "ESNext",
+    "esModuleInterop": true,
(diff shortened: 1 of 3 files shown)


react-instantsearch-hooks/query-suggestions/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
```
sarahdayan, line 1: Most of it is generated by `tsc --init`.

I added the JSX transform so we don't have to import React.

Haroenv (approved):

aymeric-giraudet (approved):

PR #424, opened 2022-11-15
fix(vue-instantsearch): bump server-side-rendering example's vue-IS

aymeric-giraudet:
Fixed another example just by bumping to the latest version of `vue-instantsearch`.

Also I forgot to commit `yarn.lock` file for the nuxt example so I'm putting it here as well.

[Fixed CodeSandbox here](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/fix/vue-ssr-bump-dependency/vue-instantsearch/server-side-rendering)

### DIFF
--- a/vue-instantsearch/server-side-rendering/src/main.js
+++ b/vue-instantsearch/server-side-rendering/src/main.js
@@ -4,6 +4,16 @@ import { createRouter } from './router';
 import { createServerRootMixin } from 'vue-instantsearch';
 import algoliasearch from 'algoliasearch/lite';
 import qs from 'qs';
+import _renderToString from 'vue-server-renderer/basic';
+
+function renderToString(app) {
+  return new Promise((resolve, reject) => {
+    _renderToString(app, (err, res) => {
+      if (err) reject(err);
+      resolve(res);
+    });
+  });
+}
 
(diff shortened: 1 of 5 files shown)


vue-instantsearch/server-side-rendering/src/main.js
```diff
@@ -4,6 +4,16 @@ import { createRouter } from './router';
...
+import _renderToString from 'vue-server-renderer/basic';
+
+function renderToString(app) {
+  return new Promise((resolve, reject) => {
+    _renderToString(app, (err, res) => {
+      if (err) reject(err);
+      resolve(res);
+    });
+  });
+}
```
sarahdayan, line 16: Looks like it wasn't just about bumping the dependency, but also updating the code. Why is this new code needed?

aymeric-giraudet: Indeed ! In the newer versions `findResultsState` is required and it throws `Error: findResultsState requires `renderToString: (component) => Promise<string>` in the first argument.` when not provided.

So we need this extra bit of code to wrap it into a promise.

Haroenv (approved):

sarahdayan (approved):

PR #430, opened 2023-01-06
nuxt3

Haroenv:


vue-instantsearch/nuxt3/pages/search.vue
```diff
@@ -0,0 +1,70 @@
...
+
+const { instantsearch } = mixin.data();
+
+provide('$_ais_ssrInstantSearchInstance', instantsearch);
+
+onServerPrefetch(async () => {
+  console.log('pre the load', this)
+  app.ssrContext.payload.algolia = await instantsearch.findResultsState({
+    // todo: what is the component? i don't see how to access the instance we currently are in
+    component: h(Search),
```
Haroenv, line 46: this isn't the component instance, it doesn't have `.render` as expected, so the temporary render done by findResultsState doesn't include the widgets

PR #446, opened 2023-03-26
Use sendEvent instead of bindEvent

kai687:
This PR updates the Insights middleware example for InstantSearch.js to use the `sendEvent` function instead of `bindEvent`.


Haroenv (changes requested):
🙆‍♂️

Haroenv (approved):

kai687:
CircleCI says it's green, but doesn't report the status properly to GitHub.

PR #452, opened 2023-05-17
chore: add codesandbox link to all code samples

dhayab:
Quick PR to add links to code sandboxes for all the code samples in this repository.

For Angular InstantSearch code samples, the links redirect to the appropriate branch.

### DIFF
--- a/angular-instantsearch/autocomplete-results-page/README.md
+++ b/angular-instantsearch/autocomplete-results-page/README.md
@@ -1,5 +1,7 @@
 # autocomplete-results-page
 
+[![Edit autocomplete-results-page](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/fix-angular-12/angular-instantsearch/autocomplete-results-page)
+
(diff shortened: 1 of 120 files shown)


angular-instantsearch/autocomplete-results-page/README.md
```diff
@@ -1,5 +1,7 @@
 # autocomplete-results-page
 
+[![Edit autocomplete-results-page](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/fix-angular-12/angular-instantsearch/autocomplete-results-page)
```
Haroenv, line 3: maybe we can make that branch master? not sure why it isn't merged

Haroenv (approved):
looks right, thanks!

angular-instantsearch/autocomplete-results-page/README.md
dhayab: If I'm remembering correctly, the version of angular in master is newer than the one in the branch, but we found out Codesandbox only works with a version set to Angular 12 for us.

Alternatively, I could link all angular-instantsearch code samples to cloud sandboxes, which will work on master at the cost of a longer cold boot time. Wdyt?

PR #454, opened 2023-07-03
Fix: Update Old CDN Links

SeanAtAlgolia:
Update CDN Links to support the new templating system introduced in version 4.46. Current versions will throw an error.

### DIFF
--- a/instantsearch.js/algolia-insights/index.html
+++ b/instantsearch.js/algolia-insights/index.html
@@ -39,8 +39,8 @@
       </div>
     </div>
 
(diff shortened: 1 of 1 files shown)


Haroenv:
Maybe this is linked to a local cache you have set up? @4 should take the latest version available

SeanAtAlgolia:
@Haroenv I was able to observe this behavior in the CodeSandBox for this repo. It throws an error for me "html is not a function". I tried clearing my browser cache but this didn't solve the problem. However it worked in an incognito window.
https://codesandbox.io/s/github/algolia/doc-code-samples/tree/master/instantsearch.js/algolia-insights?file=/index.html

PR #456, opened 2023-07-04
feat(React InstantSearch Hooks): introduce range slider example with Radix UI

sarahdayan:
This introduces a new React InstantSearch Hooks example which creates a custom `useRangeSlider` Hook and integrates it with [Radix UI's `<Slider>`](https://www.radix-ui.com/docs/primitives/components/slider) component.

https://algolia.atlassian.net/browse/FX-2419


react-instantsearch-hooks/range-slider-with-radix-ui/src/App.js
```diff
@@ -0,0 +1,84 @@
...
+              <Pagination />
+            </div>
+          </div>
+        </div>
+      </InstantSearch>
+    </div>
+  );
+}
+
+function useRangeSlider(props) {
```
Haroenv, line 49: this should be useRange IMO (which we have in the lib already, no need to rewrite it)

react-instantsearch-hooks/range-slider-with-radix-ui/src/App.js
```diff
@@ -0,0 +1,84 @@
...
+  return useConnector(connectRange, props);
+}
+
+function RangeSlider(props) {
```
Haroenv, line 53: why is it all in a single file? maybe having RangeSlider in its own file is easier for people to copy paste?

Haroenv (commented):
👍

react-instantsearch-hooks/range-slider-with-radix-ui/src/App.js
sarahdayan: Ah, I didn't realize [the current docs were outdated](https://www.algolia.com/doc/api-reference/widgets/range-slider/react-hooks/). I'll change it.

react-instantsearch-hooks/range-slider-with-radix-ui/src/App.js
Haroenv: yeah that's right, although I think the order probably could be changed, as the focus of this example is about RangeSlider, so it should be defined before the App/usage. I suggested a separate file so that can be the focused file in the codesandbox link

sarahdayan: Focusing the file from the docs is a fair point. I've reflected this in 5b50b594.

aymeric-giraudet (approved):

Haroenv (approved):

PR #464, opened 2023-08-10
fix(react-instantsearch): use → addMiddlewares

dhayab:


Haroenv (approved):
maybe we should run the code mods here too?

dhayab:
> maybe we should run the code mods here too?

Codemods won't transform files, as imports and package names are already updated, only `use` was seemingly left.

PR #476, opened 2023-09-22
feat(react-native): update getting-started and migrate query-suggestions

dhayab:
This PR updates the existing React Native example to the latest versions of React Native and Expo. It also adds here a query-suggestions code sample that was not migrated in the monorepo.

## Notable changes

Both examples:

- configure non-default highlight pre/post tags and set `useHits({ ...props, escapeHTML: false })` to have proper highlight rendering (because `getHighlightedParts()` from instantsearch.js uses hardcoded highlight delimiters instead of the placeholders)
- use `100vh` to properly set the list container's height for the react-native web platform, fixing a behaviour where the infinite list would always call `refine()` there without user interaction
- have an updated version of `Highlight.tsx` which allows specifying custom highlighted/nonHighlighted styles as props

Query suggestions is a simplified version of what we shared for React InstantSearch v6. It doesn't display images in the infinite list and doesn't refine on category, only on query.


react-instantsearch-native/getting-started/src/SearchBox.tsx
```diff
@@ -1,44 +1,54 @@
...
+  const [value, setValue] = useState(query);
   const inputRef = useRef<TextInput>(null);
 
-  function setQuery(newQuery: string) {
-    setInputValue(newQuery);
-    refine(newQuery);
-  }
+  // Track when the value coming from the React state changes to synchronize
+  // it with InstantSearch.
+  useEffect(() => {
```
Haroenv, line 16: in the React InstantSearch codebase we managed to have a searchbox without effect, is that possible here too?

Haroenv (approved):
looks solid, thanks!

react-instantsearch-native/getting-started/src/SearchBox.tsx
dhayab: Good eye, this works without them. I updated the search box in both examples.

PR #481, opened 2024-01-17
fix(render-state): use renderstate and set explicit sortby

dhayab:


instantsearch.js/render-state/src/app.js
```diff
@@ -32,11 +32,8 @@ const emptyHits = connectHits(({ hits, widgetParams }) => {
     return;
   }
 
-  const refinementListsUiState = search.getUiState()[indexName].refinementList;
-  const isPearRefined =
-    refinementListsUiState &&
-    refinementListsUiState.brand &&
-    refinementListsUiState.brand.includes('Pear');
+  const brandUiState = search.getUiState()[indexName].refinementList?.brand;
```
Haroenv, line 35: still using ui state, no?

Haroenv (commented):
not sure if this is the whole diff you wanted?

instantsearch.js/render-state/src/app.js
dhayab: 🤦‍♂️ thanks, i accidentally reapplied part of my changes.

Haroenv (approved):

PR #490, opened 2024-07-09
chore: migrate cra and vue-cli examples to vite

dhayab:
This PR is a prerequisite for the introduction of algoliasearch@5 in our code samples as:

- CRA has issues with optional chaining from the bundled script
- Vue CLI incorrectly targets the UMD bundle when importing from lite

Both issues don't exist with Vite.


Haroenv (approved):
🙌  had wanted to do this for a while, thanks!

aymeric-giraudet (approved):

PR #495, opened 2024-10-07
refactor: remove string templates

shaejaz:
This PR replaces string/hogan.js templates with `htm` templates in all demos that use them

[FX-3101]

[FX-3101]: https://algolia.atlassian.net/browse/FX-3101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/instantsearch.js/autocomplete-results-page/src/app.js
+++ b/instantsearch.js/autocomplete-results-page/src/app.js
@@ -57,13 +57,13 @@ search.addWidgets([
   instantsearch.widgets.hits({
     container: '#hits',
     templates: {
-      item: `
+      item: (hit, { html, components }) => html`
         <div>
           <header class="hit-name">
-            {{#helpers.highlight}}{ "attribute": "name" }{{/helpers.highlight}}
+            ${components.Highlight({ hit, attribute: 'name' })}
           </header>
           <p class="hit-description">
-            {{#helpers.highlight}}{ "attribute": "description" }{{/helpers.highlight}}
+            ${components.Highlight({ hit, attribute: 'description' })}
           </p>
         </div>
       `,
--- a/instantsearch.js/geo-search-tourism/src/app.js
+++ b/instantsearch.js/geo-search-tourism/src/app.js
@@ -35,23 +35,23 @@ $script(
       instantsearch.widgets.hits({
         container: '#hits',
         templates: {
-          item:
-            '<div class="hit col-sm-3">' +
-            '<div class="pictures-wrapper">' +
-            '<img class="picture" src="{{picture_url}}" />' +
-            '<img class="profile" src="{{user.user.thumbnail_url}}" />' +
-            '</div>' +
-            '<div class="infos">' +
-            '<h4 class="media-heading">{{#helpers.highlight}}{ "attribute": "name" }{{/helpers.highlight}}</h4>' +
-            '<p>' +
-            '{{room_type}} - ' +
-            '{{#helpers.highlight}}{ "attribute": "city" }{{/helpers.highlight}},' +
-            '{{#helpers.highlight}}{ "attribute": "country" }{{/helpers.highlight}}' +
-            '</p>' +
-            '</div>' +
-            '</div>',
-          empty:
-            '<div class="text-center">No results found matching <strong>{{query}}</strong>.</div>',
+          item: (hit, { html, components }) => html`
+            <div class="hit col-sm-3">
+              <div class="pictures-wrapper">
+                <img class="picture" src="${hit.picture_url}" />
+                <img class="profile" src="${hit.user.user.thumbnail_url}" />
+              </div>
+              <div class="infos">
+                <h4 class="media-heading">${components.Highlight({ hit, attribute: 'name' })}</h4>
+                <p>
+                  ${hit.room_type} - ${components.Highlight({ hit, attribute: 'city' })}, ${components.Highlight({ hit, attribute: 'country' })}
+                </p>
+              </div>
+            </div>'
+          `,
+          empty: ({ query }, { html }) => html`
+            <div class="text-center">No results found matching <strong>${query}</strong>.</div>
+          `
         },
       }),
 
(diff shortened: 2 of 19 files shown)


instantsearch.js/multi-index-hits/src/app.js
```diff
@@ -21,8 +21,9 @@ search.addWidgets([
   instantsearch.widgets.hits({
     container: '#hits-instant-search',
     templates: {
-      item:
-        '{{#helpers.highlight}}{ "attribute": "name" }{{/helpers.highlight}}',
+      item: (hit, { html, components }) => html`
+        ${components.Highlight({ hit, attribute: 'name' })}
```
Haroenv, line 25: no big deal but really you don't need to put this inside a `html` template. Fine either way for me

instantsearch.js/render-state/src/app.js
```diff
@@ -77,7 +77,7 @@ search.addWidgets([
   }),
   panel({
     templates: {
-      header: 'Brand',
+      header: () => 'Brand',
```
Haroenv, line 80: not needed but fine

Haroenv (approved):
thanks! you checked every instance of `templates` inside the js folders?

instantsearch.js/autocomplete-results-page/src/app.js
```diff
@@ -57,13 +57,13 @@ search.addWidgets([
   instantsearch.widgets.hits({
     container: '#hits',
     templates: {
-      item: `
+      item: (hit, { html, components }) => html`
         <div>
           <header class="hit-name">
-            {{#helpers.highlight}}{ "attribute": "name" }{{/helpers.highlight}}
+            ${components.Highlight({ hit, attribute: 'name' })}
```
aymeric-giraudet, line 63: I wonder if it's better to do it the intended `htm`/Preact way like : 
```
item: (hit, { html, components: { Highlight } }) => html`
  <header class="hit-name">
    <${Highlight} hit=${hit} attribute="name" />
  </header>
`
```
However we call these templates as functions anyway so I think they remount every time instead of rerendering anyway.

instantsearch.js/geo-search-tourism/src/app.js
```diff
@@ -35,23 +35,23 @@ $script(
...
-            '{{#helpers.highlight}}{ "attribute": "country" }{{/helpers.highlight}}' +
-            '</p>' +
-            '</div>' +
-            '</div>',
-          empty:
-            '<div class="text-center">No results found matching <strong>{{query}}</strong>.</div>',
+          item: (hit, { html, components }) => html`
+            <div class="hit col-sm-3">
+              <div class="pictures-wrapper">
+                <img class="picture" src="${hit.picture_url}" />
```
aymeric-giraudet, line 41: You can do `src=${hit.picture_url}` normally

instantsearch.js/refresh-cache-periodically/src/app.js
```diff
@@ -22,16 +22,16 @@ search.addWidgets([
...
+            ${components.Highlight({ hit, attribute: 'name' })}
           </header>
-          <img src="{{image}}" align="left" />
+          <img src="${hit.image}" align="left" />
           <p class="hit-description">
-            {{#helpers.highlight}}{ "attribute": "description" }{{/helpers.highlight}}
+            ${components.Highlight({ hit, attribute: 'description' })}
           </p>
-          <p class="hit-price">\${{price}}</p>
+          <p class="hit-price">\$${hit.price}</p>
```
aymeric-giraudet, line 34: I think `$` sign was escaped and we do not need the `\` anymore ?

instantsearch.js/refresh-cache-user-action/src/app.js
```diff
@@ -22,17 +22,17 @@ search.addWidgets([
...
+      item: (hit, { html, components }) => html`
+        <div>
+          <header class="hit-name">
+            ${components.Highlight({ hit, attribute: 'name' })}
+          </header>
+          <img src="${hit.image}" align="left" />
+          <p class="hit-description">
+            ${components.Highlight({ hit, attribute: 'description' })}
+          </p>
+          <p class="hit-price">\$${hit.price}</p>
```
aymeric-giraudet, line 34: Same comment as before

instantsearch.js/refresh-cache-periodically/src/app.js
shaejaz: This was for appending a '$' before the price which is just a number/float value

instantsearch.js/autocomplete-results-page/src/app.js
Haroenv: we don't document the htm/preact way yet, and the components don't have state, so I think the way it's written in this pr is good :)

instantsearch.js/refresh-cache-periodically/src/app.js
Haroenv: doesn't it still need to be escaped for a literal $?

aymeric-giraudet: Yeah but is the backslash before still needed with `html` ?

instantsearch.js/autocomplete-results-page/src/app.js
shaejaz: Yeah, might be better visually. I think in most examples/tests in instantsearch, the `components` are used like this so I kept it for consistency. Open to changing it.

instantsearch.js/render-state/src/app.js
shaejaz: Are strings still supported? I thought it only accepts a function now.

shaejaz:
> thanks! you checked every instance of `templates` inside the js folders?

Yes I think so, just updated the 'Classic HTML strings' in the dropdown as well

Haroenv (approved):
perfect, you can merge this today, no need to wait on hogan to be actually removed

